PR #26, opened 2021-03-24
New property support for lights!

### LINKED ISSUE #10: Add offset property support for lights (interior lights, corner lights, spotlights)

pandincus:
Fixes #10


EasyCommands/CommandParsers/ParameterParsers.cs
```diff
@@ -97,6 +97,12 @@ partial class Program {
         static String[] colorKeywords = { "color" };
         static String[] textKeywords = { "text", "message" };
 
+        static String[] blinkIntervalKeywords = { "blinkinterval", "blinkInterval", "interval" };
```
MerlinofMines, line 100: Given we included "interval", we might consider having the property name be "intervl" in case we can use this across other block types

MerlinofMines (approved):

PR #33, opened 2021-04-01
Revamped ScriptTests by adding ScriptTest utility class

pandincus:
Revamped ScriptTests by adding ScriptTest utility class
This greatly reduces the required boilerplate in writing script tests
by delegating most of that to the construction of a ScriptTest instance.

ScriptTest authors only need to provide the actual script,
the mocked objects, and the verifications/assertions for those mocks.

As part of this, I replaced block and group entity providers with
mocking of the GridTerminalSystem. However, I haven't yet configured
it to work properly for display blocks. I wanted to get this bit out
and reviewed before then proceeding to address the display block gap.

As a result, you'll see changes to EasyCommands for the removal of the
block provider delegates, as those are no longer required.

Other notable changes:
* Changed STATE in Program to be public so we can both examine it
  and set it from tests
* Changed the BroadcastMessageProvider delegate to not be static
  so that we don't have to potentially reset it after each test run

### DIFF
--- a/EasyCommands.Tests/ScriptTests/ScriptTest.cs
+++ b/EasyCommands.Tests/ScriptTests/ScriptTest.cs
@@ -1,0 +1,182 @@
+﻿using IngameScript;
+using Malware.MDKUtilities;
+using Moq;
+using Sandbox.ModAPI.Ingame;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using static IngameScript.Program;
+
+namespace EasyCommands.Tests.ScriptTests
+{
+    /// <summary>
+    /// Represents an instance of a ScriptTest. Constructing a new ScriptTest is akin to setting
+    /// up the test, as it instantiates required objects and performs basic setup.
+    /// 
+    /// This implements <see cref="IDisposable"/>, so any post-test cleanup is performed
+    /// upon destruction of the class. Please use this inside of a using statement.
+    /// </summary>
+    class ScriptTest : IDisposable
+    {
+        private GenericListDictionary mockedBlocksByName;
+        private List<IMyBlockGroup> mockedGroups;
+        private Mock<IMyGridTerminalSystem> gridMock;
+        private Program program;
+
+        public List<String> Logger { get; private set; }
+
+        public ScriptTest(String script)
+        {
+            mockedBlocksByName = new GenericListDictionary();
+            mockedGroups = new List<IMyBlockGroup>();
+            Logger = new List<String>();
+
+            // Init static fields that may have been modified from other tests, prepare for testing
+            Program.STATE = ProgramState.STOPPED;
+            Program.LOG_LEVEL = Program.LogLevel.SCRIPT_ONLY;
+            // Setup the CUSTOM_DATA to return the given script
+            // And other required config for mocking
+            gridMock = new Mock<IMyGridTerminalSystem>();
+            var me = new Mock<IMyProgrammableBlock>();
+            MDKFactory.ProgramConfig config = default;
+            config.GridTerminalSystem = gridMock.Object;
+            config.ProgrammableBlock = me.Object;
+            config.Echo = (message) => Logger.Add(message);
+            program = MDKFactory.CreateProgram<Program>(config);
+            me.Setup(b => b.CustomData).Returns(script);
+
+            // Default behavior for broadcast messages
+            // TODO: Replace this with mock objects passed to config setup in Program
+            program.broadcastMessageProvider = () => new List<MyIGCMessage>();
+
+            // Default behavior for locating blocks on the grid
+            gridMock.Setup(g => g.GetBlockGroups(It.IsAny<List<IMyBlockGroup>>(), It.IsAny<Func<IMyBlockGroup, bool>>()))
+                .Callback((Action<List<IMyBlockGroup>, Func<IMyBlockGroup, bool>>)GetBlockGroupsCallback);
+
+            // TODO: Handle custom logic in TextSurfaceHandler
+        }
+
+        /// <summary>
+        /// Run the script until the program reaches the specified state, or until a limited
+        /// number of runs has been reached. The default number of runs before exit is 100.
+        /// </summary>
+        /// <param name="state">The desired end-state.</param>
+        /// <param name="runLimit">The limited number of runs. Defaults to 100</param>
+        public void RunUntil(ProgramState state, int runLimit = 100)
+        {
+            int runCounter = 0;
+            do
+            {
+                MDKFactory.Run(program);
+                runCounter++;
+            } while (Program.STATE != state && runCounter < runLimit);
+        }
+
+        /// <summary>
+        /// Given a custom name for a block (or set of blocks), set up the custom names
+        /// for the given mocked blocks and store them for later retrieval when requested by
+        /// the grid terminal system.
+        /// 
+        /// Since the GridTerminalSystem's GetBlocksOfType method doesn't return values but
+        /// instead modifies the given list, this uses Moq's Callback functionality to
+        /// invoke a callback that will attempt to retrieve the stored blocks.
+        /// </summary>
+        /// <typeparam name="T">The type of the blocks we're mocking retrieval for.</typeparam>
+        /// <param name="name">The custom name of the block(s) we're mocking.</param>
+        /// <param name="blockMocks">Blocked mocks to be stored and used later.</param>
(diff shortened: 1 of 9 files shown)


EasyCommands.Tests/ScriptTests/ScriptTest.cs
```diff
@@ -0,0 +1,182 @@
...
+            // TODO: Handle custom logic in TextSurfaceHandler
+        }
+
+        /// <summary>
+        /// Run the script until the program reaches the specified state, or until a limited
+        /// number of runs has been reached. The default number of runs before exit is 100.
+        /// </summary>
+        /// <param name="state">The desired end-state.</param>
+        /// <param name="runLimit">The limited number of runs. Defaults to 100</param>
+        public void RunUntil(ProgramState state, int runLimit = 100)
```
MerlinofMines, line 67: Nice!

EasyCommands.Tests/ScriptTests/ScriptTest.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            // Store the mock block objects by their type
+            mockedBlocksByName.Add(typeof(T), blockMocks.Select(b => b.Object).ToList());
+        }
+
+        public class GenericListDictionary
+        {
+            private Dictionary<Type, List<Object>> dictionary = new Dictionary<Type, List<object>>();
+
+            public void Add<T>(Type type, List<T> values)
```
MerlinofMines, line 166: I think you can simplify and remove the type parameter by doing:

```
public void Add<T>(List<T> values) {
  dictionary.Add(typeof(T), values.Select(x => (Object)x).ToList());
}
```

EasyCommands.Tests/ScriptTests/ScriptTest.cs
```diff
@@ -0,0 +1,182 @@
...
+            {
+                dictionary.Add(type, values.Select(x => (Object)x).ToList());
```
MerlinofMines, line 168: I think this will fail if you try to add to the dictionary twice.

Instead, I think you need to fetch the current list by key (adding an empty list if not exists) and then appending the objects to the given list.  That way we can mock multiple blocks of the same type with different custom names and/or belonging to different groups.

EasyCommands.Tests/ScriptTests/ScriptTest.cs
MerlinofMines, line 67: Thoughts on passing a delegate "checker" to this instead of using runLimit (you could still pass runLimit == 100 as the default)?

It'd be kind of nice to say "run until X condition", sort of like breakpoints in the script to then check assertions of the state of the world.

EasyCommands.Tests/ScriptTests/ScriptTest.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <param name="blockMocks">The blocks being mocked and that will be returned later.</param>
+        public void MockBlocksInGroup<T>(String groupName, params Mock<T>[] blockMocks) where T : class, IMyTerminalBlock
+        {
+            var mockGroup = new Mock<IMyBlockGroup>();
+            mockGroup.Setup(g => g.Name).Returns(groupName);
+            mockGroup.Setup(g => g.GetBlocksOfType(It.IsAny<List<T>>(), It.IsAny<Func<T, bool>>()))
+                .Callback((Action<List<T>, Func<T, bool>>)GetBlocksOfTypeCallback);
+            mockedGroups.Add(mockGroup.Object);
+
+            SetupMockBlocksByType(blockMocks);
```
MerlinofMines, line 116: I'm trying to think how this will work if I have a script where I want to get the same mocked blocks by name, and by being in group.  I think right now this would double-add to mockedBlocksByName.  

If we change GenericListDictionary to `Dictionary<Type,Set<object.>>` does that alleviate?

Alternatively we could separate adding mock blocks and then adding those blocks to groups, but i like the one shot method.  Thoughts?

EasyCommands.Tests/ScriptTests/ScriptTest.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        public ScriptTest(String script)
+        {
+            mockedBlocksByName = new GenericListDictionary();
+            mockedGroups = new List<IMyBlockGroup>();
+            Logger = new List<String>();
+
+            // Init static fields that may have been modified from other tests, prepare for testing
+            Program.STATE = ProgramState.STOPPED;
+            Program.LOG_LEVEL = Program.LogLevel.SCRIPT_ONLY;
```
MerlinofMines, line 38: nit: we might want to put these after the MDKFactor.CreateProgram call, lest we start moving some of this setup into the Program constructor

EasyCommands/BlockHandlers/BlockHandlers.cs
```diff
@@ -10,9 +10,6 @@
 namespace IngameScript {
     partial class Program {
         public static class BlockHandlerRegistry {
-            public delegate List<Object> BlockProvider(BlockType blockType, String name);
-            public static BlockProvider BLOCK_PROVIDER = GetBlocks;
-            public static BlockProvider GROUP_BLOCK_PROVIDER = GetBlocksInGroup;
```
MerlinofMines, line 13: nice

EasyCommands.Tests/ScriptTests/ScriptTest.cs
```diff
@@ -0,0 +1,182 @@
...
+        private void GetBlockGroupsCallback(List<IMyBlockGroup> groups, Func<IMyBlockGroup, bool> collect)
+        {
+            // Invoke the given collector function on all groups and return those groups matching
+            groups.AddRange(mockedGroups
+                .Select(x => x)
+                .Where(collect != null ? collect : (x) => true)
+                .ToList());
+        }
+
+        public void Dispose()
```
MerlinofMines, line 137: Do we need this one?

nit: I think default is LogLevel.INFO.  I think this also hints at doing the defaulting in the Program constructor, and then overriding in the our test setup after creating the program. 

Thoughts?

MerlinofMines (changes requested):
Looks awesome!  Trying out the "request changes" functionality with a couple things I wants your thoughts on.

Cheers!

EasyCommands.Tests/ScriptTests/ScriptTest.cs
pandincus: Nice! Missed this. Done.

EasyCommands.Tests/ScriptTests/ScriptTest.cs
pandincus: That's right! I wanted to write a test that exercised a variety of scenarios like multiple blocks of the same type. Ran out of steam the other night but I've added that now (and addressed this issue).

EasyCommands.Tests/ScriptTests/ScriptTest.cs
pandincus: Done! The predicate currently accepts the ScriptTest as its input and I figure we can expose aspects of the program through properties on the ScriptTest.

EasyCommands.Tests/ScriptTests/ScriptTest.cs
pandincus: This is a great catch. Yup! We would double-add.

Changing it to a Set should do the trick here. I've also added a test explicitly around manipulating blocks by-themselves and as a group. Check it out!

EasyCommands.Tests/ScriptTests/ScriptTest.cs
pandincus: Yeah, I moved 'em for now, then we can take another pass at changing these to instance variables.

EasyCommands/BlockHandlers/BlockHandlers.cs
pandincus: Felt good to delete it 😄

EasyCommands.Tests/ScriptTests/ScriptTest.cs
pandincus: Oh for sure, these should all be instance variables. The Dispose will be useful for anything we can't change to an instance variable, or any other post-execution cleanup we want to do. FYI, the `Dispose` method is invoked automatically when we close out the `using` statement.

> nit: I think default is LogLevel.INFO.

Yup! That's why I'm setting it to that in Dispose.

Anyway, you're totally right on this better being done with instance variables, but I didn't want to make a bunch of changes to your state management variables in this CR. I can do that in a follow-up, though!

PR #35, opened 2021-04-05
Refactored ParameterProcessors to use Rule Based Processing

MerlinofMines:
This commit significantly reduces expensive boilerplate code for ParameterProcessors by moving almost all processors to use a RuleProcessor and delegates.  This has
dramatically reduced the file and character size of ParameterProcessors.

I also added some missing tests for Iteration and Control command parameter processing.

I also removed some now unneeded constructors for some command types.

I also removed "Resume" as a specific control type and changed "Start" to act like Resume, as it had no purpose previously (was the same as Restart which didn't
quite make sense anyway).

Lastly, I changed the ParameterProcessor ranking to be set specifically be the list ordering instead of being inferred by type, as this was a problem for multiple
rule based processors that need to process the same type but maintain priority ordering for correctness.

### DIFF
--- a/EasyCommands/CommandParsers/ParameterProcessors.cs
+++ b/EasyCommands/CommandParsers/ParameterProcessors.cs
@@ -24,59 +24,262 @@ namespace IngameScript {
 
             private static List<ParameterProcessor> parameterProcessors = new List<ParameterProcessor>
             {
-                  new ParenthesisProcessor(),
-                  new FunctionProcessor(),
-                  new AssignmentProcessor(),
-                  new RunArgumentProcessor(),
-                  new SelfSelectorProcessor(),
-                  new SelectorProcessor(),
-                  new VariableSelectorProcessor(),
-                  new PrimitiveProcessor(),
-                  new RedundantComparisonProcessor(),
-                  new UniOperationProcessor(),
-                  new MultiplyProcessor(),
-                  new AddProcessor(),
-                  new AndProcessor(),
-                  new OrProcessor(),
-                  new NotProcessor(),
-                  new VariableComparisonProcessor(),
-                  new BlockComparisonProcessor(),
-                  new IndexProcessor(),
-                  new ConditionalSelectorProcessor(),
-                  new IndexSelectorProcessor(),
-                  new PropertyAggregationProcessor(),
-                  new AggregationProcessor(),
-                  new IteratorProcessor(),
-                  new IfProcessor(),
-                  new IterationProcessor(),
-                  new ActionProcessor(),
-                  new PrintCommandProcessor(),
-                  new WaitProcessor(),
-                  new FunctionCallCommandProcessor(),
-                  new VariableAssignmentProcesor(),
-                  new SendCommandProcessor(),
-                  new ListenCommandProcessor(),
-                  new ControlProcessor(),
-                  new IterationProcessor(),
-                  new ConditionalCommandProcessor(),
-                  new QueueCommandProcessor()
+                new ParenthesisProcessor(),
+
+                //FunctionProcessor
+                OneValueRule<FunctionCommandParameter,StringCommandParameter>(
+                    requiredRight<StringCommandParameter>(),
+                    (p,name) => {
+                        FunctionDefinition definition;
+                        if(!FUNCTIONS.TryGetValue(name.GetValue().Value, out definition)) throw new Exception("Unknown function: " + name.GetValue().Value);
+                        return new FunctionDefinitionCommandParameter(p.Value, definition);
+                    }),
+
+                //AssignmentProcessor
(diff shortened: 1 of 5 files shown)


MerlinofMines:
I'm deploying the script with full minimization enabled to see if/how this reduced the final script file size.  I'll update once I have the results

MerlinofMines:
Official minified characters is now just under 75000, at 74991!  

Not the 10,000 I was originally hoping to save, but I'll take it! :)

EasyCommands/CommandParsers/ParameterProcessors.cs
```diff
@@ -116,1146 +319,307 @@ public static class ParameterProcessorRegistry {
...
-                variable = null;
+        public abstract class ParameterProcessor<T> : ParameterProcessor where T : class, CommandParameter {
+            int rank;
+            public virtual List<Type> GetProcessedTypes() { return new List<Type>() { typeof(T) }; }
+            public virtual int GetRank() => rank;
+            public virtual void SetRank(int r) => rank = r;
+            public bool CanProcess(CommandParameter p) {
+                return p is T;
             }
+            public abstract bool Process(List<CommandParameter> p, int i, out List<CommandParameter> finalParameters);
```
pandincus, line 475: With rank living in `ParameterProcessor`, you could implement `IComparable<ParameterProcessor>` and have:

```c#
public int CompareTo(ParameterProcessor other)
{
    return Rank.compareTo(other.Rank);
}
```

Then, you can delete your custom insertion logic up above and just do `parameterProcessors.sort()` to keep it sorted. (The `sort` method, by default, looks to see if the objects in the list implement `IComparable` or `IComparer`, and if so, uses its given comparison method to perform the sorting)

EasyCommands/CommandParsers/ParameterProcessors.cs
```diff
@@ -116,1146 +319,307 @@ public static class ParameterProcessorRegistry {
...
             }
 
             static void InsertProcessor(ParameterProcessor processor, List<ParameterProcessor> parameterProcessors) {
-                int priority = priorityByParameterProcessor[processor.GetType()];
+                int priority = processor.GetRank();
                 int i = 0;
-                while (i < parameterProcessors.Count() && priorityByParameterProcessor[parameterProcessors[i].GetType()] < priority) i++;
+                while (i < parameterProcessors.Count() && parameterProcessors[i].GetRank() < priority) i++;
                 parameterProcessors.Insert(i, processor);
             }
```
pandincus, line 332: Related to my other comment on taking advantage of rank living in `ParameterProcessor` and `IComparable`, I think you can get rid of this method and just add the appropriate processor to your list in `AddProcessors`. Then, at the end of your addition/searching logic, call `sort()` on it and it should be good to go. Should save a few more lines! (And should be a bit more efficient since you're not having to iterate over the list inside every iteration of `AddProcessors`).

EasyCommands/CommandParsers/ParameterProcessors.cs
```diff
@@ -116,1146 +319,307 @@ public static class ParameterProcessorRegistry {
...
-                return new CommandReferenceParameter(new PrintCommand(variable));
-            }
-
-            public override void Initialize() {
-                variable = null;
+        public abstract class ParameterProcessor<T> : ParameterProcessor where T : class, CommandParameter {
+            int rank;
+            public virtual List<Type> GetProcessedTypes() { return new List<Type>() { typeof(T) }; }
+            public virtual int GetRank() => rank;
+            public virtual void SetRank(int r) => rank = r;
```
pandincus, line 471: Pretty sure `Rank` can be a property. `public virtual int Rank { get; set; }`

pandincus (approved):

EasyCommands/CommandParsers/ParameterProcessors.cs
MerlinofMines: Great call out, I've made ParameterProcessor implement IComparable.

I'm struggling on how to add attributes to the ParameterProcessor interface.  Making the interface an abstract class adds more characters (need to specify "public", "abstract", "override" keywords which can't be minified)  

Maybe you've got some ideas on how to simplify?

For now I condensed as much as I could but get & set methods are still present on the interface. :(

EasyCommands/CommandParsers/ParameterProcessors.cs
MerlinofMines: yep, i cleaned this up

EasyCommands/CommandParsers/ParameterProcessors.cs
MerlinofMines: can't use virtual on the interface, but i was able to condense it down pretty far! :) Basically have to copy the definition (plus public) in the abstract class

PR #40, opened 2021-04-05
Fixed Explicit Strings not working for Assignment and Function Calls

### LINKED ISSUE #27: Bug: Cannot assign variable to explicit string
assign 'myLocation' to avg "Fighter Cockpit" location

fails to parse.

however,

assign "myLocation" to avg "Fighter Cockpit" location

works

Need to fix and add a parameter processing test to verify

MerlinofMines:
This fix, built on top of pull #35 , fixes an outstanding bug where you can't assign variables or call functions using an explicit string.

Specifically, this commit fixes #27
This commit also moves all the processor tests to the correct namespace and moves the variable assignment tests into SimpleVariableParamterProcessorTests

### DIFF
--- a/EasyCommands.Tests/ParameterParsingTests/AggregateBlockPropertyProcessorTests.cs
+++ b/EasyCommands.Tests/ParameterParsingTests/AggregateBlockPropertyProcessorTests.cs
@@ -2,7 +2,7 @@
 using System;
 using static IngameScript.Program;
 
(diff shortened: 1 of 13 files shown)


MerlinofMines:
This is a re-submitted pull request from #36, will close the other one.  Merging since I have approval from previous and no changes have been made

Issue #12, opened 2021-03-22
Global "All" selector


MerlinofMines:
It'd be great to have some sort of "All" Selector so that you could get or set properties for all blocks of the given type.  This would be nice for turning off all the lights when you have multiple light groups, for example.

"all" is currently an aggregation mode.  we *could* try to have an aggregationMode=(any/all)  Selector => "All Selector" processor invoked after we have tried conditional block parsing?

MerlinofMines:
Working on this.

Little snag.  "all" is already taken by AggregationMode.

So, I can parse AggregationMode followed by BlockType as Selector.  Or need to think of a better word for this?

MerlinofMines:
Should be easier to add a new rule for this with the new rule processors

PR #43, opened 2021-04-08
Added "all" and implicit selector support

MerlinofMines:
This change allows you to perform an action on all blocks of a given type without defining a group.

The new selectors can be used in block commands to perform actions, or in aggregate block conditions to check for block state.

You can also use this to get aggregate block properties.

This commit implements #12

### DIFF
--- a/EasyCommands/BlockHandlers/BlockHandlers.cs
+++ b/EasyCommands/BlockHandlers/BlockHandlers.cs
@@ -50,9 +50,15 @@ namespace IngameScript {
                 if (!blockHandlers.ContainsKey(blockType)) throw new Exception("Unsupported Block Type: " + blockType);
                 return blockHandlers[blockType];
             }
+
+            public static List<Object> GetAllBlocks(BlockType blockType) {
+                return blockHandlers[blockType].GetAllBlocks();
+            }
+
             public static List<Object> GetBlocks(BlockType blockType, String customName) {
                 return blockHandlers[blockType].GetBlocks(customName);
             }
+
             public static List<Object> GetBlocksInGroup(BlockType blockType, String groupName) {
                 return blockHandlers[blockType].GetBlocksInGroup(groupName);
             }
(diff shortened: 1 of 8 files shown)


EasyCommands/BlockHandlers/BlockHandlers.cs
```diff
@@ -238,6 +245,12 @@ public abstract class TerminalBlockHandler<T> : BlockHandler<T> where T : class,
...
 
+            public override List<T> GetAllBlocksOfType() {
+                List<T> blocks = new List<T>();
+                PROGRAM.GridTerminalSystem.GetBlocksOfType<T>(blocks);
+                return blocks;
+            }
+
             public override List<T> GetBlocksOfType(String name) {
                 List<T> blocks = new List<T>();
                 PROGRAM.GridTerminalSystem.GetBlocksOfType<T>(blocks, block => block.CustomName.Equals(name));
```
pandincus, line 256: These two methods are identical with the exception of the predicate passed to `PROGRAM.GridTerminalSystem.GetBlocksOfType<T>(...)`. How about we have the new method call the old one with a null name, and the old method interprets null name in order to not pass a condition? Or, simply pass a condition that returns true to save characters.

```c#
PROGRAM.GridTerminalSystem.GetBlocksOfType<T>(blocks,
    block => name != null
        ? block.CustomName.Equals(name)
        : true);
```

EasyCommands/BlockHandlers/TextSurfaceHandlers.cs
```diff
@@ -30,6 +30,15 @@ public class TextSurfaceHandler : BlockHandler<IMyTextSurface> {
...
+                PROGRAM.GridTerminalSystem.GetBlocksOfType(blocks);
+
+                List<IMyTextSurface> surfaces = new List<IMyTextSurface>();
+                blocks.ForEach((b) => Add(b, surfaces));
+                return surfaces;
+            }
+
             public override List<IMyTextSurface> GetBlocksOfType(String name) {
                 List<IMyTerminalBlock> blocks = new List<IMyTerminalBlock>();
                 PROGRAM.GridTerminalSystem.GetBlocksOfType(blocks, block => block.CustomName.Equals(name));
```
pandincus, line 44: Similar comment as in `TerminalBlockHandler` re: similar functions.

pandincus (approved):
Only one actionable comment, but can be looked at after shipping as well.

EasyCommands/BlockHandlers/BlockHandlers.cs
MerlinofMines: Thanks for keeping me honest!  I agree I tried and gave up too early, will address

PR #44, opened 2021-04-08
Fixed Bug with Rotor angle not honoring direction

### LINKED ISSUE #14: Rotor SetDirection is ignoring Direction
Currently Rotors ignore direction when you request to set a rotor to an angle.

So "turn the rotor to 30 counterclockwise" will still move clockwise to 30 if the rotor determines that it's faster to move clockwise to get to 30.

Also note, "30 degrees" doesn't work currently. Need to think about this separately.

MerlinofMines:
This commit mostly fixes the issue with rotor angle not obeying directions.  It turns out there are some cases where the game itself won't properly wrap around.
I added tests for the cases where obeying direction should now work.

I found multiple bugs introduced by previous changes as part of this which also had to be fixed, including:
* Subtraction Operation was not working correctly (was adding)
* SetPropertyDirection was ignoring direction for all blocks, not just rotors
* Increment/Decrement commands were broken as the relative command parameter was passed to BlockCommand twice.

This commit fixes #14

### DIFF
--- a/EasyCommands.Tests/ScriptTests/RotorBlockTests.cs
+++ b/EasyCommands.Tests/ScriptTests/RotorBlockTests.cs
@@ -1,0 +1,200 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using Moq;
+using Sandbox.ModAPI.Ingame;
+using VRageMath;
+
+namespace EasyCommands.Tests.ScriptTests {
+    [TestClass]
+    public class RotorBlockTests {
+        [TestMethod]
+        public void setRotorAngleGreaterThanMovesClockwise() {
+            String script = @"
+rotate the ""rotor"" to 30
+";
+
+            using (ScriptTest test = new ScriptTest(script)) {
+                var mockRotor = new Mock<IMyMotorStator>();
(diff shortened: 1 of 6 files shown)


EasyCommands.Tests/ScriptTests/RotorBlockTests.cs
```diff
@@ -0,0 +1,200 @@
...
+            String script = @"
+rotate the ""rotor"" clockwise by 30
+";
+
+            using (ScriptTest test = new ScriptTest(script)) {
+                var mockRotor = new Mock<IMyMotorStator>();
+                test.MockBlocksOfType("rotor", mockRotor);
+
+                mockRotor.Setup(r => r.TargetVelocityRPM).Returns(1);
+                mockRotor.Setup(r => r.Angle).Returns(345 * (float)Math.PI / 180);
```
pandincus, line 78: Oh wow, the rotor angle is represented in radians???

pandincus (approved):

EasyCommands.Tests/ScriptTests/RotorBlockTests.cs
MerlinofMines: ya..

PR #48, opened 2021-04-10
Migrated RocketVolley to EasyCommands

### LINKED ISSUE #22: Convert RocketVolley to EasyCommand Script
I'm pretty sure we can convert RocketVolley to an EasyCommands Script now.

Need to add count support! 

Here's my first attempt:

:setup
assign rocketGroup to 'My Rockets'
assign "i" to 0
assign "firing" to false
assign fireTickInterval to 8

:toggle
if {firing}
  goto "stopFiring"
else 
  goto "startFiring"

:startFiring
assign "firing" to true
async if {i} >= count of [rocketGroup] rockets assign "i" to 0
tell [rocketGroup] rockets @ {i} to shoot
wait {fireTickInterval}
assign "i" to {i} + 1
restart

:stopFiring
assign "i" to 0
assign "firing" to false

MerlinofMines:
This commit also fixes a bug with missing "not" processing for BlockCommands.

It also adds a few more keywords which should map to the GunBlockHandler.

Lastly, it adds another utility to mock expected actions in MockEntityUtility and sets the default function parse amount for tests to 1000, so that tests always
parse in 1 turn.

This commit resolves #22


EasyCommands.Tests/ScriptTests/MockEntityUtility.cs
```diff
@@ -3,12 +3,23 @@
...
 using Sandbox.ModAPI.Ingame;
+using Sandbox.ModAPI.Interfaces;
 using VRageMath;
 
 namespace EasyCommands.Tests.ScriptTests {
     class MockEntityUtility {
 
+        public static Mock<ITerminalAction> MockCalledAction<T>(Mock<T> mockBlock, String actionName) where T : class, IMyTerminalBlock {
+            Mock<ITerminalAction> terminalAction = new Mock<ITerminalAction>();
+            terminalAction.Setup(action => action.Apply(mockBlock.Object));
```
pandincus, line 16: Why call setup for the `Apply` method without a corresponding Returns or other behavior?

EasyCommands.Tests/ScriptTests/RocketVolleyTests.cs
```diff
@@ -0,0 +1,145 @@
...
+                var rocket1On = MockCalledAction(mockRocket1, "Shoot_On");
+                var rocket2Off = MockCalledAction(mockRocket2, "Shoot_Off");
+                var rocket2On = MockCalledAction(mockRocket2, "Shoot_On");
+
+                test.MockBlocksInGroup("Rockets", mockRocket1, mockRocket2);
+
+                test.RunOnce(); //Setup but no firing yet
+
+                rocket1Off.Verify(b => b.Apply(mockRocket1.Object));
+                rocket2Off.Verify(b => b.Apply(mockRocket2.Object));
```
pandincus, line 59: Huh! Neat. I haven't seen much with `ITerminalAction` yet, so I'm assuming that this is the only way we can actually verify that we're invoking these "functions" on these blocks. (Inconsistent space engineers APIs as opposed to the "actions" being methods on the blocks themselves?)

pandincus (approved):

EasyCommands.Tests/ScriptTests/RocketVolleyTests.cs
MerlinofMines: Right, some functionality they do not have APIs for, there is a generic "applyAction(name)" used instead.  You'll see this in a few block handlers.

EasyCommands.Tests/ScriptTests/MockEntityUtility.cs
MerlinofMines: I think it was failing before I added the setup.  I can re-check

MerlinofMines: Nope! Not needed. removed

PR #53, opened 2021-04-11
Added Thread Local Variable Support

### LINKED ISSUE #51: Thread Local Variables
Currently all variables are global, meaning it is difficult to have multiple threads running the same functions on different entities, since any assigned variables will start to clobber each other.

I think a decent solution to this is to have variables be thread specific by default.  Only if/when a new "global" field is specified do we add a variable to our global variables list.

I suggest "global" as an optional assignment parameter (kind of like group) so that assign + bind still operate as expected.

MerlinofMines:
This commit adds thread Local Variable Support and default s variable assignment to use thread local variables.

Additionally, this commit adds support for assigning global variables using the "global" keyword.

This commit resolves #51

### DIFF
--- a/EasyCommands.Tests/ScriptTests/MultiThreadingTests.cs
+++ b/EasyCommands.Tests/ScriptTests/MultiThreadingTests.cs
@@ -100,9 +100,6 @@ print 'Async'
 ";
 
             using (var test = new ScriptTest(script)) {
-                test.RunOnce();//Parse
-                Assert.AreEqual(0, test.Logger.Count);
-
                 test.RunOnce();//Execute main, add async threads and execute wait in all 3
 
                 Assert.AreEqual(1, test.Logger.Count);
@@ -130,9 +127,6 @@ print 'Async'
 ";
 
             using (var test = new ScriptTest(script)) {
-                test.RunOnce();//Parse
-                Assert.AreEqual(0, test.Logger.Count);
-
                 test.RunOnce();//Execute main, add async thread
 
                 Assert.AreEqual(1, test.Logger.Count);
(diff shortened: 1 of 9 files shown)


EasyCommands.Tests/ScriptTests/MultiThreadingTests.cs
```diff
@@ -165,9 +159,6 @@ public class MultiThreadingTests {
 ";
 
             using (var test = new ScriptTest(script)) {
-                test.RunIterations(2);//Parse
-                Assert.AreEqual(0, test.Logger.Count);
```
pandincus, line 169: nice

EasyCommands.Tests/ScriptTests/ScriptTest.cs
```diff
@@ -43,6 +43,7 @@ public ScriptTest(String script)
             mockedGroups = new List<IMyBlockGroup>();
             Logger = new List<String>();
             RunCounter = 0;
+            Program.FUNCTION_PARSE_AMOUNT = 1000;
```
pandincus, line 46: what's the default? Should we set it back in `Dispose`?

EasyCommands/Program.cs
```diff
@@ -58,6 +59,21 @@ public partial class Program : MyGridProgram {
...
+        public Variable GetVariable(String variableName) {
+            Thread currentThread = GetCurrentThread();
+            if(currentThread.threadVariables.ContainsKey(variableName)) {
+                return currentThread.threadVariables[variableName];
+            } else if (globalVariables.ContainsKey(variableName)) {
+                return globalVariables[variableName];
+            } else {
+                throw new Exception("No Variable Exists for name: " + variableName);
+            }
+        }
```
pandincus, line 75: Nice!

pandincus (approved):

PR #54, opened 2021-04-11
Added Upper/Lower Limit Support

### LINKED ISSUE #50: Limit Property Support
We have several blocks that would benefit from a "limit" property, especially when looking at direction limits such as piston/rotor limits.

I suggest a new "limit" property, as well as add "upper" and "lower" as directional keywords

MerlinofMines:
This commit adds the "range" property to piston & rotor blocks using a new directional property handler to get the lower/upper limit by direction type.

This commit also adds tests for the new behavior for both rotors and pistons.

Lastly, this commit moves "lower" from an action word to a simple direction word, which should be backwards compatible.

This commit resolves #50


EasyCommands/BlockHandlers/PistonBlockHandlers.cs
```diff
@@ -40,6 +41,36 @@ public class PistonHeightHandler : SimpleNumericPropertyHandler<IMyPistonBase> {
             }
         }
 
+        static float GetLimit(IMyPistonBase piston, DirectionType direction) {
+            switch (direction) {
+                case DirectionType.UP:
+                case DirectionType.CLOCKWISE:
```
pandincus, line 47: clockwise for a piston?

pandincus (approved):

EasyCommands/BlockHandlers/PistonBlockHandlers.cs
MerlinofMines: ya maybe we don't need that one

PR #55, opened 2021-04-14
Added TextSurface and Subclass Mocking Support

### LINKED ISSUE #46: Fix Mocking to Support Inheritance
Currently our mock blocks doesn't consider that blocks might be a subclass of the one requested and does not return the mocks.

We should change how we resolve blocks by type to iterate through and filter instead of using a type map.

MerlinofMines:
This commit adds support for mocking text surfaces on text surface providers.  It also adds tests confirming we can now test textsurfaces.

Additionally, this commit replaces our mock block grid and mock block groups with a proper implementation to overcome the issues with subclass casting which the
moq framework can't really deal with.

This commit resolves #46

### DIFF
--- a/EasyCommands.Tests/ScriptTests/ScriptTest.cs
+++ b/EasyCommands.Tests/ScriptTests/ScriptTest.cs
@@ -21,9 +21,7 @@ namespace EasyCommands.Tests.ScriptTests
     /// </summary>
     class ScriptTest : IDisposable
     {
-        private GenericListDictionary mockedBlocksByName;
-        private List<IMyBlockGroup> mockedGroups;
-        private Mock<IMyGridTerminalSystem> gridMock;
+        MockGridTerminalSystem mockGrid;
         private Program program;
         private Mock<IMyProgrammableBlock> me;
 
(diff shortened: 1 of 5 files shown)


EasyCommands.Tests/ScriptTests/ScriptTest.cs
```diff
@@ -191,48 +158,123 @@ public void Dispose()
...
+
+            public string Name => name;
+
+            public void AddBlock(IMyTerminalBlock block) {
+                mockBlocks.Add(block);
+            }
+
+            public void AddBlocks(List<IMyTerminalBlock> blocks) {
+                blocks.ForEach(block => mockBlocks.Add(block));
+            }
```
pandincus, line 191: `mockBlocks.unionWith(blocks)` should do this.

EasyCommands.Tests/ScriptTests/ScriptTest.cs
```diff
@@ -191,48 +158,123 @@ public void Dispose()
...
+            public void GetBlocksOfType<T>(List<T> blocks, Func<T, bool> collect = null) where T : class {
+                blocks.AddRange(mockBlocks
+                    .Where(block => block is T)
+                    .Select(block => (T) block)
+                    .Where(block => collect == null ? true : collect(block))
+                    .ToList());
             }
         }
 
+        public class MockGridTerminalSystem : IMyGridTerminalSystem {
```
pandincus, line 215: Honestly, probably the right way to go here.

pandincus (approved):

EasyCommands.Tests/ScriptTests/ScriptTest.cs
MerlinofMines: it doesn't quite.  i tried it and decided it's more characters and less efficient to do it.

What you actually have to do is:

`mockBlocks = new HashSet<IMyTerminalBlock>(blocks.Union(blocks));`

vs existing

`blocks.ForEach(block => mockBlocks.Add(block));`

PR #57, opened 2021-04-16
This commit adds the ability to process multiple possible parsings fo…

### LINKED ISSUE #32: Implicit Aggregate Property Support
Currently you have to specify "avg" to indicate that you are trying to assign a variable, or parse a value, from a block(s) property.

Example:

assign a to my average location

This sounds silly.  As a user assume I should be able to say:

assign a to my location

The problem (currently) is that this conflicts with BlockCommand, which technically can parse

"my location" as a BlockCommand, attempting to set the "location" property of the "my" selector to Bool true.

So..some changes are needed to support this.

I think we might be able to say that actions need to indicate "do an action" using an ActionParameter like 
"tell", "set", "rotate", "move", "raise".  UPDATE: Nope.  "trigger the alarm" is a valid statement and clearly a BlockCommand.  

If this was the case, then we could define an "ImplicitAggregateProperty" processor, which runs after BlockCommandProcessor, to take Selector+property+direction and convert to AggregatePropertyVariable, assuming "avg" as the AggregationMode.

This would enable the above use case.

MerlinofMines:
This commit adds the ability to process multiple possible parsings for ambiguous requests such as selector + property, where it could either be an aggregate variable or a block command depending on the context.

It then adds a new branch processor for block + property + direction in isolation which default parses as an implicit block aggregate property, with the fallback branch being a stand alone block command.

Implicit Aggregate Property uses a new AggregationMode for "value".  This aggregation mode throws an exception if the input selector has more than 1 value.  The
result is that you will get an exception during execution if you try to use an implicit aggregation on a selector property that contains more than one block.

Also adds test and modifies some of the existing script tests to confirm the new parsing is working.

I also verified the changes by updating my "follow me" ship locally and it works!

This commit resolves #32

### DIFF
--- a/EasyCommands/CommandParsers/ParameterProcessors.cs
+++ b/EasyCommands/CommandParsers/ParameterProcessors.cs
@@ -327,7 +327,7 @@ namespace IngameScript {
             /// This method inline processes the given list of command parameters.
             /// Any ambiguous parsing branches which were found during processing are also returned as additional entries.
             /// If the desired result (typically a command) does not result from the returned parse, the returned
-            /// branches can be re-processed to see if a correct parse results from the altneratve branch.
+            /// branches can be re-processed to see if a correct parse results from the alternate branches.
             /// This can continue until no alternate branches are returned.
             /// </summary>
             /// <param name="commandParameters"></param>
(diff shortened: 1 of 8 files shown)


EasyCommands/CommandParsers/ParameterProcessors.cs
```diff
@@ -311,11 +323,21 @@ public static class ParameterProcessorRegistry {
                 }
             }
 
-            public static void Process(List<CommandParameter> commandParameters) {
+            /// <summary>
+            /// This method inline processes the given list of command parameters.
+            /// Any ambiguous parsing branches which were found during processing are also returned as additional entries.
+            /// If the desired result (typically a command) does not result from the returned parse, the returned
+            /// branches can be re-processed to see if a correct parse results from the altneratve branch.
```
pandincus, line 330: `alt[e]rnat[i]ve`

pandincus (approved):
Dude, this is great.

EasyCommands/CommandParsers/ParameterProcessors.cs
MerlinofMines: whoops! fixed

PR #61, opened 2021-04-18
Added Support for Double Quote Explicit Strings

### LINKED ISSUE #60: Implicit Variable from Double Quotes Support
Right now a double quoted string is always converted to a Selector if any of the words happen to be a block type or block group type.  As such, you have to remember to use single quotes when using a string containing these types of words.

The below example fails to parse because [print] [selector] cannot be parsed
`
if [myLights] are on
  Print "The light are on"
`

We should use our new branching capability to add another possible parse for the string into an explicit string, which should alleviate this.

MerlinofMines:
This commit uses the new branching support to add new parsing support for parsing double quoted strings containing block type tokens.

Before this commit, a command like assign "v" to "my piston" would fail as "my piston" is interpreted as a selector and then the command will not parse.
This commit introduced a new branch to interpret all double quoted strings as an explicit string.  The result is that the above command now parses.

This has some ramifications to Implicite Aggregate Properties, as we must distinguish between an explicit string and an implicite aggregate property in this case.
The following rules, implemented by this CR, disambiguate:

1.  If the aggregator is specified (count, sum, avg) then property is optional as there is no ambiguity.
2.  If an implicit aggregator is intended, property must be supplied, i.e. assign v to "test piston" height presumes to get the height property of "test piston"
3.  If only a string is provided without an aggregator or a property, then we do not attempt to convert to an aggregate property.  If parsing the string as a
selector leads to a successful parse (aggregate condition, etc) then great.  Otherwise the only other branch considered will be that the string must have intended to
be a literal string.

This commit adds a simple test showing the new support works (Rule 3).  The existing tests all passing is confirmation of the above rules working as intended.

This commit resolves #60

### DIFF
--- a/EasyCommands/CommandParsers/ParameterProcessors.cs
+++ b/EasyCommands/CommandParsers/ParameterProcessors.cs
@@ -56,18 +56,20 @@ namespace IngameScript {
                     (p, blockType) => new SelectorCommandParameter(new SelfEntityProvider(blockType.HasValue() ? blockType.GetValue().Value : BlockType.PROGRAM))),
 
                 //SelectorProcessor
-                TwoValueRule<StringCommandParameter,BlockTypeCommandParameter,GroupCommandParameter>(
-                      optionalRight<BlockTypeCommandParameter>(),optionalRight<GroupCommandParameter>(),
-                      (p,blockType,group) => {
-                        if (!blockType.HasValue()) {
-                            BlockTypeCommandParameter type = findLast<BlockTypeCommandParameter>(p.SubTokens);
-                            if (type != null) blockType.SetValue(type);
-                            GroupCommandParameter g = findFirst<GroupCommandParameter>(p.SubTokens);
-                            if (g != null) group.SetValue(g);
-                        }
-                        return blockType.HasValue();
-                      },
-                      (p,blockType,group) => new SelectorCommandParameter(new SelectorEntityProvider(blockType.GetValue().Value, group.HasValue(), new StaticVariable(new StringPrimitive(p.Value))))),
+                new BranchingProcessor<StringCommandParameter>(
+                    TwoValueRule<StringCommandParameter,BlockTypeCommandParameter,GroupCommandParameter>(
+                          optionalRight<BlockTypeCommandParameter>(),optionalRight<GroupCommandParameter>(),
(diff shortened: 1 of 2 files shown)


EasyCommands/CommandParsers/ParameterProcessors.cs
```diff
@@ -211,10 +213,11 @@ public static class ParameterProcessorRegistry {
                 //AmgiguousSelectorPropertyProcessor
                 new BranchingProcessor<SelectorCommandParameter>(
                     TwoValueRule<SelectorCommandParameter,PropertyCommandParameter,DirectionCommandParameter>(
-                        optionalEither<PropertyCommandParameter>(), optionalEither<DirectionCommandParameter>(),
-                        (s,p,d) => new VariableCommandParameter(new AggregatePropertyVariable(PropertyAggregatorType.VALUE, s.Value, p.HasValue() ? p.GetValue().Value : (PropertyType?)null, d.HasValue() ? d.GetValue().Value : (DirectionType?)null))),
+                        requiredEither<PropertyCommandParameter>(), optionalEither<DirectionCommandParameter>(),
```
MerlinofMines, line 216: This change enforces Rule 2.  Implicit Aggregate Properties must specify the requested property.

EasyCommands/CommandParsers/ParameterProcessors.cs
```diff
@@ -211,10 +213,11 @@ public static class ParameterProcessorRegistry {
...
+                        (s,p,d) => new VariableCommandParameter(new AggregatePropertyVariable(PropertyAggregatorType.VALUE, s.Value, p.GetValue().Value, d.HasValue() ? d.GetValue().Value : (DirectionType?)null))),
                     TwoValueRule<SelectorCommandParameter,PropertyCommandParameter,DirectionCommandParameter>(
                         optionalEither<PropertyCommandParameter>(), optionalEither<DirectionCommandParameter>(),
+                        (s,p,d) => p.HasValue() || d.HasValue(),//Must have at least one!
```
MerlinofMines, line 220: This was a small bugfix.  BlockCommand cannot parse (fails with exception) if you try to give it a selector without property OR direction, yet the rules allowed it.

pandincus (approved):

Issue #13, opened 2021-03-22
Debug logging for new Parameter Processors


MerlinofMines:
The new method of parameter processing doesn't have debug logs, so it's tough to understand why something did or did not parse.

I think something in the top level function for what was processed and the output type could be useful.  we could also consider logging as part of each parameter processor.  Maybe in SimpleParameterProcessor?

MerlinofMines:
Alright, taking a look at this one.

MerlinofMines:
tricky tricky.

Im thinking that a way to get or set strings/tokens per command parameter type is probably best here.

trying to figure out how I can add this support without adding a bunch of boilerplate to CommandParameters

PR #64, opened 2021-04-21
Parameter processor debugging enhancement

MerlinofMines:
This commit implements much better processor debugging support with 2 commits:

1 - Add the parse tree to the debug output.  When LogLevel is set to Debug, the output will contain detailed information about how the command was parsed.

2 - Add line numbers to the exception output when a command cannot be parsed.  

This commit implements #13

### DIFF
--- a/EasyCommands/CommandParsers/ParameterProcessors.cs
+++ b/EasyCommands/CommandParsers/ParameterProcessors.cs
@@ -31,29 +31,29 @@ namespace IngameScript {
                     requiredRight<StringCommandParameter>(),
                     (p,name) => {
                         FunctionDefinition definition;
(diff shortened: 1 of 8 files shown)


EasyCommands/CommandParsers/ParameterProcessors.cs
```diff
@@ -356,7 +358,11 @@ public static class ParameterProcessorRegistry {
                             } else revisit = true;
                         }
                     }
-                    if (processed) { processorIndex = 0; continue; }
+                    if (processed) {
+                        Debug(String.Join(" ", commandParameters.Select(p => CommandParameterToString(p)).ToList()));
+                        processorIndex = 0;
```
pandincus, line 363: Let's add a script test for this. Don't bother asserting every line, just pick one.

pandincus (approved):

PR #78, opened 2021-05-14
Added support for assigning variables using the variable syntax {}

MerlinofMines:
I've seen a few times (and done this myself) where a user tries to say:

"assign {a} to {b}" since normally you use {} or []to interact with a variable.

Currently you have to do either:
assign "a" to {b}
assign 'a' to {b}

Now you can also do
assign {a} to {b}
assign [a] to {b}

### DIFF
--- a/EasyCommands/CommandParsers/ParameterProcessors.cs
+++ b/EasyCommands/CommandParsers/ParameterProcessors.cs
@@ -45,6 +45,13 @@ namespace IngameScript {
             TwoValueRule<AssignmentCommandParameter,GlobalCommandParameter,ExplicitStringCommandParameter>(
                 optionalRight<GlobalCommandParameter>(), requiredRight<ExplicitStringCommandParameter>(),
                 (p,g,name) => new VariableAssignmentCommandParameter(name.GetValue().value, p.useReference, g.HasValue())),
+            TwoValueRule<AssignmentCommandParameter,GlobalCommandParameter,VariableCommandParameter>(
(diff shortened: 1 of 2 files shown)


MerlinofMines:
Should we consider doing VariableSelector [a] as well?

EasyCommands/CommandParsers/ParameterProcessors.cs
```diff
@@ -45,6 +45,10 @@ partial class Program : MyGridProgram {
             TwoValueRule<AssignmentCommandParameter,GlobalCommandParameter,ExplicitStringCommandParameter>(
                 optionalRight<GlobalCommandParameter>(), requiredRight<ExplicitStringCommandParameter>(),
                 (p,g,name) => new VariableAssignmentCommandParameter(name.GetValue().value, p.useReference, g.HasValue())),
+            TwoValueRule<AssignmentCommandParameter,GlobalCommandParameter,VariableCommandParameter>(
```
MerlinofMines, line 48: Should we consider doing VariableSelector [a] as well?

EasyCommands/CommandParsers/ParameterProcessors.cs
```diff
@@ -45,6 +45,10 @@ partial class Program : MyGridProgram {
...
+                optionalRight<GlobalCommandParameter>(), requiredRight<VariableCommandParameter>(),
+                (p,g,name) => name.GetValue().value is InMemoryVariable,
```
MerlinofMines, line 50: null check needed

PR #92, opened 2021-06-01
Add missing block handlers

MerlinofMines:
This pull requests adds several missing blockhandlers & properties, including

* collectors
* ejectors
* connector strength
* laser antenna
* jumpdrive
* decoy
* engine / wind turbine 

It also separates hinges from rotors so that the two can be operated on independently.


MerlinofMines:
This latest change, on top of the production CR, puts us at 90522 characters.  Getting close to that 100,000 wall!

PR #93, opened 2021-06-11
Collection support enhancement

MerlinofMines:
Finally, Collection Support is out.

This pull request adds collection support to EasyCommands, enabling the use of Lists of variables.

Functionality includes:
*assign variable to list
*assign list index to value
*retrieval of list values by index
*aggregation of list values, including count, sum, avg, min, max, etc.
*aggregate condition of lists, including all values >, =, < a given value (effectively contains)
*append
*insert
*insert range
*sublist (effectively remove and remove range)
*multi-dimensional list support

### DIFF
--- a/EasyCommands/CommandParsers/ParameterParsers.cs
+++ b/EasyCommands/CommandParsers/ParameterParsers.cs
@@ -22,7 +22,7 @@ namespace IngameScript {
         //Internal (Don't touch!)
         Dictionary<String, List<CommandParameter>> propertyWords = new Dictionary<string, List<CommandParameter>>();
 
-        string[] separateTokens = new[] { "(", ")", "[", "]", ",", "+", "*", "/", "!", "^", ".." }; //TODO: Fix '-' but make sure "-3" still works
+        string[] separateTokens = new[] { "(", ")", "[", "]", ",", "+", "*", "/", "!", "^", ".." };
 
         public void InitializeParsers() {
             //Ignored words that have no command parameters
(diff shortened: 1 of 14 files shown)


EasyCommands/Common/Variables.cs
```diff
@@ -195,5 +203,81 @@ public class InMemoryVariable : Variable {
...
+            public Variable expectedList;
+            public Variable index;
+
+            public ListIndexVariable(Variable list, Variable i) {
+                expectedList = list;
+                index = i;
+            }
+
+            //TODO: Add Lookup by string support (Dictionary)
+            //TODO: Add List support!  If index is list then return list containing at all requested indexes.  if empty, return expectedList
```
MerlinofMines, line 229: This comment is actually addressed

EasyCommands/CommandParsers/ParameterParsers.cs
```diff
@@ -20,11 +20,13 @@
 namespace IngameScript {
     partial class Program {
         //Internal (Don't touch!)
-        private Dictionary<String, List<CommandParameter>> propertyWords = new Dictionary<string, List<CommandParameter>>();
+        Dictionary<String, List<CommandParameter>> propertyWords = new Dictionary<string, List<CommandParameter>>();
+
+        string[] separateTokens = new[] { "(", ")", "[", "]", ",", "+", "*", "/", "!", "^", ".." };
```
pandincus, line 25: we don't need one for modulo?

pandincus (approved):
this is amazing

EasyCommands/CommandParsers/ParameterParsers.cs
MerlinofMines: good call out! Ya we can add modulo :D

Issue #94, opened 2021-06-11
Dictionary Support


MerlinofMines:
Adding onto collection support, we should figure out how to store and retrieve string indexed values

assign a["myString"] to "thing" -> add index to end of a with string value "myString" and whose value is "thing"

assign myValue to a["myString"]

how will contains work?  I suppose it will reference key index values.

When printing how will we know that "key" = "value" ? 

Do we want to support dictionary instantiation?  If so, what is syntax?  Maybe easier to instantiate an empty list? Or..can we claim something like "=" ?  assign a to ["me" = 1, "too" = 2] ?

Maybe we can say "=" -> map assignment and "==" is comparison?

How does that affect adding string values to list?  

Lots of stuff to think about

MerlinofMines:
We could also consider taking back the {} syntax and re-use for map key/value definitions: {a,"value"}
assign a{"myThing"} to "newThing"

PR #95, opened 2021-06-12
Dictionary support enhancement

MerlinofMines:
This commit adds on to #93 by adding dictionary support.  Collections has been pulled into it's own primitive type which stores a list of optionally keyed values.

You can get or set values in a list by specifying a string variable instead of a numeric variable as the index.

You can also get a keyed sublist by asking for multiple string keys. 

I also added support for 2 new operations on lists, namely "keys" and "values", for retrieving just the keys or just the values from a keyed list (similar to dictionary).  Non-keyed values will not be returned by "keys", but all values are returned by "values".

Merging keyed lists de-dupes string based index values, keeping the only the keyed values from the 2nd list but otherwise maintaining the order.

It also adds new support for removing items from lists either by index (numeric) or by key (string).  Multiple items can be removed at once using lists, and numeric/string removal can be combined

This pull request implements #94


MerlinofMines:
This latest enhancement puts us at 93205 chars.

PR #102, opened 2021-11-04
Dynamic Property Support

### LINKED ISSUE #19: Dynamic specifier support
We've added [a] to represent "treat variable a as a selector"

Let's do more

How about 'thing[a]'?

property[a] => treat a as a property
direction[a] => treat a as a direction
type[a] => treat as a block type.  
constant[string] => We can implement constants like PI, E, etc.

MerlinofMines:
Whew, this one was quite a doozy!

At long last, this pull request resolves #19.  Well, sort of.

It does not enable dynamic support for all types of entities (directions, etc).  

What it does do is enable you to specify properties dynamically though, for example:
```set the "test lights" "limit" property to 10```

Another handy addition is the ability to control properties using the SpaceEngineers PropertyIds.  This means that EasyCommands can now get & set all attributes of any TerminalBlock as long as that attribute is retrievable via a PropertyId.  This also means that any TerminalBlock added by 3rd party mods can now also be controlled via it's properties.  

So why the huge code refactor?

Well... a couple of challenges.

First, I was tired of seeing "if direction == null getProperty else getPropertyDirection".  This code was sprinkled across several places and it just was...well..ugly.

Second, I was tired of doing inconsistent property resolution using directions in some places, primitives in others, etc.  There was an opportunity to consolidate this all in a single place and I took it!  

Third (and most importantly) there's an issue with retrieving and setting some properties by propertyId.  I wanted the ability to say something like 
``` set the "test door" "close" property to true ``` 

But there's a catch.  Some properties mean "the inverse of the actual property".

For example, "close" = open false.  

So..if you want to say:
```print "Closed: " + "test door" "closed" property```

What you really mean is "get the open property, and tell me it's opposite value".  And conversely,
```set the "test door" "closed" property to true```

means "set the test door open property to not true"

Both of these scenarios mean that the value that needs to be retrieved or set has to be inversed before it is returned or stored.  This meant that the same place where we resolve the actual property needed the ability to change the value associated with the property being manipulated.

To support all of the above, I needed to take Direction & Value and push them into the PropertySupplier construct, so that this single entity would now have all the information about what was property was being manipulated and with what values.

There are a couple useful side-effects of this, to be explored at a future date.  First, I was able to simplify and remove several interface methods to BlockHandler.  Instead of "Set, SetDirection, Move" there's now just "Update".  Instead of "Increment,IncrementDirection" there's not just "Increment".  Second, PropertyHandlers now have a lot more flexibility in how they choose to handle a request to get or set a property, as they can inspect the PropertySupplier to look for the existence of a direction (or not).  This hasn't been leveraged yet, but the possibility is there and will be useful for future work such as supporting directional limits, where the thing to do is dependent on whether a direction was passed and if so which one.

Some limitations:
Dynamic Property support does not extend to Value Properties or directions.

So you can't do either of the following
``` set the "test cargo" '"gold ingot" amount' to 5 ```
``` set the "test rover" "backwards velocity" property to 5

These could theoretically be added in the future, but I didn't see the need  yet.


MerlinofMines:
Merging!

PR #128, opened 2022-01-06
Force locale to en-us in unit tests

### LINKED ISSUE #127: Tests are locale dependant
Right now most of the test that involve parsing a floating point number or comparing strings containing floating point numbers
fail if the default locale of the system does __not__ have `.` as decimal seperator like in french or german.

jgersti:
This PR fixes #127

### DIFF
--- a/EasyCommands.Tests/ParameterParsingTests/AggregateBlockPropertyProcessorTests.cs
+++ b/EasyCommands.Tests/ParameterParsingTests/AggregateBlockPropertyProcessorTests.cs
@@ -1,6 +1,5 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
-using System.Globalization;
 using System.Collections.Generic;
 using Malware.MDKUtilities;
 using IngameScript;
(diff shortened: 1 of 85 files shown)


EasyCommands.Tests/ParameterParsingTests/AggregateBlockPropertyProcessorTests.cs
```diff
@@ -10,6 +11,11 @@ namespace EasyCommands.Tests.ParameterParsingTests {
     public class AggregateBlockPropertyProcessorTests {
         List<object> aggregationList = new List<object> { 1, 2, 3 };
 
+        [TestInitialize]
+        public void InitializeTestClass() {
+              System.Threading.Thread.CurrentThread.CurrentCulture = new CultureInfo("en-US");      
+        }
+
```
MerlinofMines, line 18: Making this change in all current (and future) tests is pretty heavy handed.  Is there a way to set this globally so that it's the case for all tests, in a way that doesn't have to be maintained for all future tests?

jgersti: It might be possible to hide the TestInitialize method in an abstract base. 
One would still need to remeber to derive future test from this base but the cognitive burden should be lower.

jgersti:
You might want to squash the branch before merging/rebasing.

MerlinofMines (approved):
Thanks for making the adjustments!

PR #131, opened 2022-01-08
Fix cast operators

### LINKED ISSUE #121: Some cast operators are broken
Some cast operators are broken.

Trying to cast anything into a number or a color gives a parsing error.
Casting a number to vector yields `cannot convert number to color`. (From what i can tell from the source converting a number to a color is allowed.)

Also the cast from bool to number would behave inconsistent, if it would work. `bool -> number -> bool` always yields `false`.

The other casts may exhibit some unexpected behaviour aswell.

jgersti:
This PR is a starting point to fix #121.

- [x] Add tests for all possible casts 
    - 36 tests for the current expected behaviour
    - test for the proposed behaviour for `number as bool` 
    the 2 variants of the `number as bool` test are mutually exclusive!
- [x] fix mixed up messages for invalid casts
- [x] fix InvalidCastException for `color as vector`
    One of the casts in the sequence `Vector3 -> Object -> Vector3D` throws the exception 
- [x] fix parsing errors
- [x] clean up the implementation

edit:
- moved the remainig task (behavioral changes) into a seperate issue/PR #139
- added task

### DIFF
--- a/EasyCommands/CommandParsers/ParameterParsers.cs
+++ b/EasyCommands/CommandParsers/ParameterParsers.cs
@@ -102,7 +102,8 @@ namespace IngameScript {
             AddPropertyWords(Words("close", "closed", "shut"), Property.OPEN, false);
             AddPropertyWords(PluralWords("font"), Property.FONT);
             AddPropertyWords(PluralWords("text", "message", "argument"), Property.TEXT);
-            AddPropertyWords(PluralWords("color", "foreground"), Property.COLOR);
+            AddPropertyWords(AllWords(Words("colors"), PluralWords("foreground")), Property.COLOR);
+            AddAmbiguousWords(Words("color"), new PropertyCommandParameter(Property.COLOR));
             AddPropertyWords(PluralWords("background"), Property.BACKGROUND);
             AddPropertyWords(Words("power", "powered"), Property.POWER);
             AddPropertyWords(Words("enable", "enabled", "arm", "armed"), Property.ENABLE);
@@ -185,9 +186,11 @@ namespace IngameScript {
             AddWords(Words("average", "avg"), new PropertyAggregationCommandParameter((blocks, primitiveSupplier) => SumAggregator(blocks, primitiveSupplier).Divide(ResolvePrimitive(Math.Max(1, blocks.Count())))));
             AddWords(Words("minimum", "min"), new PropertyAggregationCommandParameter((blocks, primitiveSupplier) => blocks.Select(b => primitiveSupplier(b)).DefaultIfEmpty(ResolvePrimitive(0)).Aggregate((a, b) => (a.Compare(b) < 0 ? a : b))));
             AddWords(Words("maximum", "max"), new PropertyAggregationCommandParameter((blocks, primitiveSupplier) => blocks.Select(b => primitiveSupplier(b)).DefaultIfEmpty(ResolvePrimitive(0)).Aggregate((a, b) => (a.Compare(b) > 0 ? a : b))));
-            AddWords(Words("count", "number"), new PropertyAggregationCommandParameter((blocks, primitiveSupplier) => ResolvePrimitive(blocks.Count())));
+            AddWords(Words("count"), new PropertyAggregationCommandParameter((blocks, primitiveSupplier) => ResolvePrimitive(blocks.Count())));
+            AddAmbiguousWords(Words("number"), new PropertyAggregationCommandParameter((blocks, primitiveSupplier) => ResolvePrimitive(blocks.Count())));
             AddWords(Words("sum", "total"), new PropertyAggregationCommandParameter(SumAggregator));
-            AddWords(Words("list", "collection"), new PropertyAggregationCommandParameter((blocks, primitiveSupplier) => ResolvePrimitive(new KeyedList(blocks.Select(b => new StaticVariable(primitiveSupplier(b))).ToArray()))));
+            AddWords(Words("collection"), new PropertyAggregationCommandParameter((blocks, primitiveSupplier) => ResolvePrimitive(new KeyedList(blocks.Select(b => new StaticVariable(primitiveSupplier(b))).ToArray()))));
+            AddAmbiguousWords(Words("list"), new PropertyAggregationCommandParameter((blocks, primitiveSupplier) => ResolvePrimitive(new KeyedList(blocks.Select(b => new StaticVariable(primitiveSupplier(b))).ToArray()))));
 
             //Operations Words
             AddWords(Words("("), new OpenParenthesisCommandParameter());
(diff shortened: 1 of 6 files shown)


MerlinofMines:
Great start on this and thank you so much for adding a new test file specifically for comparisons, I agree this is needed.

The painful bit here is going to be around comparison.

So right now, comparison for the equals operator is done as "a.Compare(b) == 0", not unreasonably.

The Compare() functions needs some work, I think.

Currently it performs the COMPARE operation on the given types, casts to a number, and returns it.

I think we simply need to fix this.

Calling COMPARE is still fine, I think.   But we need to look at the result.  If the result is a boolean (returnType=Return.Bool) return 0.  If numeric, return the numeric directly?

That'll get rid of the bad Numeric->Bool hack that's in place. 

Thoughts?

jgersti:
I would leave the comparison stuff for later. The main concern should be the parsing errors.

Is there way to reasonably way to step through the parser to see where the cast branch is not matched/taken ?

MerlinofMines:
I'm not sure you'll be able to put all of the cast functions in the right place without fixing comparison, but dig through it and see what you find.

As for walking through, you can debug the tests to see what's happening.  I had to re-add the debug configurations to get it to work though (shouldn't have gotten rid of them previously).  Re-added in the #135  PR.  Feel free to take a look.  I will likely push them soon.

A good point to put a breakpoint would be the Cast<T> function in Primitives on line 101.

jgersti:
All cast to `color`, `list` and `number` are broken on a fundamental level.
The token `color` is matched as a `PropertyCommandParameter`, the tokens `list` and `number` as `PropertyAggregationCommandParameter`, but the rule processor tries to match a `VariableCommandParameter`.
`PropertyAggregationCommandParameter`s also cannot be distinguished.

The only way to i see to fix this is something along the way of #132.

MerlinofMines:
The cast operation is intended to use a string as the 2nd operand.  I agree it'd be nicer if you didn't have to surround in quotes.

I suggest we get the cast operations in a good state, using "number", "list", "color" and then can look at how we might extend this.

The idea of "this word might mean 2 things" could be generalized so we don't have to create classes for each no ambiguous items, if we can find a common place in the parsing tree to put the branching.  But let's solve that as a follow up?

jgersti:
Can you take a look at the way the ambigious tokens are handled.
Unfortunatly it is rather spedific to  the casts.

The only test still failing is the expected `bool -> number` one with the changed behaviour

MerlinofMines:
I like where you are going with this, but I agree it's a little specific to casting.  Let me take a thorough look and propose something a bit more generic.

My gut says something like:

AddAmbiguousPropertyWords(Words("color"), new PropertyCommandParameter(Property.Color));

We can create a new AmbiguousPropertyCommandParameter that has a CommandParameter attribute.  

AddAmbiguousPropertyWords creates an AmbiguousStringCommandParameter whose sub CommandParameter attribute is the one you defined.  

The parameter processor (which will need to be moved above operations I think) can then look for AmbiguousStringCommandParameter and do the branching.

With this approach you shouldn't need the alternative words mapping, I believe

Thoughts?

jgersti:
move the remainig task (behavioral changes) into a seperate issue/PR #139

jgersti:
Take a look please if that was what you had in mind.

It should atleast be cleaner and more extensible than the previous approach.

.editorconfig
```diff
@@ -2,3 +2,14 @@
...
+csharp_new_line_before_members_in_object_initializers=false
+csharp_new_line_before_members_in_anonymous_types=false
+csharp_new_line_between_query_expression_clauses=false
+insert_final_newline=true
+csharp_new_line_before_else=true
+
+[*.{cs,vb}]
+indent_size=4
+tab_width=4
+end_of_line=lf
```
MerlinofMines, line 15: Can you remove these from the CR?

.gitignore
```diff
@@ -345,4 +345,5 @@ ASALocalRun/
 healthchecksdb
 
 # lib/nuget
-lib/
\ No newline at end of file
+lib/
+/EasyCommands_Backup_2022-01-07-144211726.zip
```
MerlinofMines, line 349: Let's remove this, looks like you can delete locally

EasyCommands/CommandParsers/ParameterParsers.cs
```diff
@@ -364,6 +368,22 @@ partial class Program {
...
+        {
+            AddAmbiguousWords(words, new CommandReferenceParameter(new ControlCommand { controlFunction = function }));
+        }
+
+        void AddAmbiguousPropertyWords(String[] words, Property property, bool nonNegative = true) {
+            if (!nonNegative) AddAmbiguousWords(words, new PropertyCommandParameter(property), new BooleanCommandParameter(false));
+            else AddAmbiguousWords(words, new PropertyCommandParameter(property));
+        }
+
+        void AddAmbiguousWords(String[] words, params CommandParameter[] commands) {
```
MerlinofMines, line 381: nit: "commands" -> "commandParameters"

EasyCommands/CommandParsers/ParameterParsers.cs
```diff
@@ -364,6 +368,22 @@ partial class Program {
...
+        void AddAmbigiuousControlWords(String[] words, ControlFunction function)
+        {
+            AddAmbiguousWords(words, new CommandReferenceParameter(new ControlCommand { controlFunction = function }));
+        }
+
+        void AddAmbiguousPropertyWords(String[] words, Property property, bool nonNegative = true) {
+            if (!nonNegative) AddAmbiguousWords(words, new PropertyCommandParameter(property), new BooleanCommandParameter(false));
+            else AddAmbiguousWords(words, new PropertyCommandParameter(property));
+        }
+
```
MerlinofMines, line 380: I think these functions may not be necessary yet?  If we just use AddAmbiguousWords, does that save characters?

.editorconfig
jgersti: these damn files creep in all the time ...

.gitignore
jgersti: see above ...

EasyCommands/CommandParsers/ParameterParsers.cs
jgersti: `AddAmbiguousControlWords(...)` is atleast right now completly unnecessary.
`AddAmbiguousPropertyWords(...)` only purpose right now is to makes the definitions of the tokens look the same.

MerlinofMines: let's condense for now to save chars and add the utility functions once it's less chars to use the utility vs not having it.  This is the pattern I try to follow elsewhere so that we aren't constantly hitting the character limit and having to refactor stuff to stay below 100000.

EasyCommands/CommandParsers/ParameterParsers.cs
jgersti: see below

MerlinofMines (approved):

Issue #124, opened 2022-01-04
FR: Unary Minus


jgersti:
The negation operator (unary -) is missing.
While the workaround is simple, the bahaviour is nonetheless unexpected.

MerlinofMines:
Yep, this one's been painful.

Right now "-" is only a binary operator, unless it's in front of a number (in which case it's parsed as a negative number).

I think the way to fix this is to create a new CommandParameter for "Minus".

Then, in the parsing tree, we need a branching processor to convert it to "Unary Minus" or "Binary Minus".  The branching logic will then identify whether it was indented to be unary or binary based on the surrounding context.

Shouldn't be too bad to implement.  we'll have to figure out exactly where to put the branch in the parsing tree to be most effective / least effective.

MerlinofMines:
We might change the "negative" parsing logic too, and have "-" always be separated out and treated as a unary operator.

The unary operator we can use should be "Reverse", I believe?  Whichever one currently does the ".Not()" on the value.  Can't remember exactly which one atm.

PR #132, opened 2022-01-09
Minus Operation Support

MerlinofMines:
This commit adds support for "-" to mean either subtraction or negation, depending on it's surrounding context.  This effectively enables support for negating
variables without requiring -1 * variable.  Now you can simply do -variable.  Negation takes precedence, so -a-a = (-a) - a, vs -(a-a).

This commit also changes parentheses processing to throw out branches that do not resolve to a single value.  The new expectation is that anything in parentheses
should resolve to a single value, be it a selector, variable or command.


MerlinofMines:
This commit resolves #124

PR #133, opened 2022-01-09
Fixed Power/Enable Properties

### LINKED ISSUE #114: Fix "Power" vs "Enabled" issues
Right now "Power" is the default property on a lot of blocks, where really it should be Enabled.  Many blocks override "Power" but don't really intend this.

There are also locks of blocks overriding the default boolean property with bad properties, we should revisit all of them.

Docs need to be updated to fix "Enabled" vs "Power" listed as default, and to get rid of "turn on/off" for Power on some of them that override the Power property.  We should get rid of "turn on/off" for Power property since it is only Enabled (or the default boolean property if correctly overridden).

MerlinofMines:
This commit fixes Power/Enable properties and removed several incorrect boolean default properties, including air vents, assemblers, batteries, cockpits, cameras,
gas tanks, parachutes, remote control blocks, and sensors

Also added missing "stockpiling" keywords.

Resolves #114

### DIFF
--- a/EasyCommands/BlockHandlers/ShipControllerHandlers.cs
+++ b/EasyCommands/BlockHandlers/ShipControllerHandlers.cs
@@ -34,10 +34,9 @@ namespace IngameScript {
                         return new KeyedList(waypoints.Select(w => new KeyedVariable(GetStaticVariable(w.Name), GetStaticVariable(w.Coords))).ToArray());
                     },
                     SetWaypoints);
-
+                defaultPropertiesByPrimitive[Return.BOOLEAN] = Property.AUTO;
                 defaultPropertiesByPrimitive[Return.VECTOR] = Property.TARGET;
                 defaultPropertiesByPrimitive[Return.LIST] = Property.WAYPOINTS;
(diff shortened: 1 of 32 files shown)


EasyCommands/BlockHandlers/ShipControllerHandlers.cs
```diff
@@ -83,7 +84,7 @@ public class ShipControllerHandler<T> : TerminalBlockHandler<T> where T : class,
                     TypeHandler(NumericHandler(b => -b.RollIndicator), Direction.COUNTERCLOCKWISE),
                     TypeHandler(NumericHandler(b => b.RollIndicator), Direction.CLOCKWISE));
 
-                defaultPropertiesByPrimitive[Return.BOOLEAN] = Property.ENABLE;
+                defaultPropertiesByPrimitive[Return.BOOLEAN] = Property.POWER;
```
MerlinofMines, line 87: should actually leave as enable

MerlinofMines:
Squashing to a single commit for fixing up these properties

Issue #116, opened 2022-01-02
Programmable Block Improvements


MerlinofMines:
Just a few useful properties to be adjusted for Programmable BLocks.

* Add "programmable" keyword to block type.
* Add "argument, arguments" word for Text property
* Fix Programmable Block "run" w/ boolean to use default terminal argument instead of ""

MerlinofMines:
Taking a look

PR #134, opened 2022-01-09
Programmable Block Improvements

MerlinofMines:
This commit adds "programmable" as a block word for Programmable Blocks, adds "argument" as a text property keyword, and defaults "Run" to use the default terminal
argument instead of ""


MerlinofMines:
Resolves #116

MerlinofMines:
Missing docs!

PR #138, opened 2022-01-10
Gyro Block Improvements

### LINKED ISSUE #108: GyroScope Handler Improvements
The Gyroscope needs a little love.

Currently overrides use a % from 0-1.  However SE maps those values to weird values.  For example , "0.5" = 4.77 RPM but "1.0" = 60 RPM.  I verified that if setting the property *directly* it maps 1-1, so "5" = 5 RPM.  This is a much better user experience, so we should change the definition of gyro overrides to numeric instead of percentage and change the internals appropriately.

The Power property for Gyroscopes should be "GyroPower", not enabled.  Default boolean property 
should be updated to "Enabled", not Power.

"Rotation" should get and set a vector reflecting (X:Y:Z) when no direction is passed.  

Overrides should be adjusted to be typed based.  If bool, then turn on/off.  If vector, same as Rotation.  If direction, expect numeric.  Honestly I think we can combine all of them to work the same.  That way "turn off the gyro rotation" does exactly what you want it to do, as well.

Don't forget to update docs! :)

MerlinofMines:
This commit fixes Gyro Overrides support to propertly use RPM, vs a weird % that didn't really work right.

It also extends "overrides" support to include vector, numeric and boolean.

Also, the Power property now controls Gyro power.

This PR resolves #108

### DIFF
--- a/EasyCommands/BlockHandlers/GyroscopeBlockHandlers.cs
+++ b/EasyCommands/BlockHandlers/GyroscopeBlockHandlers.cs
@@ -19,23 +19,45 @@ using VRageMath;
 
 namespace IngameScript {
     partial class Program {
-        public class GyroscopeBlockHandler : FunctionalBlockHandler<IMyGyro> {
+        public class GyroscopeBlockHandler<T> : FunctionalBlockHandler<T> where T : class, IMyGyro {
             public GyroscopeBlockHandler() {
+                var powerHandler = NumericHandler(b => b.GyroPower, (b, v) => b.GyroPower = v, 0.1f);
+                AddPropertyHandler(Property.RANGE, powerHandler);
+                AddPropertyHandler(Property.POWER, powerHandler);
+
                 AddBooleanHandler(Property.AUTO, b => !b.GyroOverride, (b, v) => b.GyroOverride = !v);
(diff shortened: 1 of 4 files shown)


EasyCommands/BlockHandlers/GyroscopeBlockHandlers.cs
```diff
@@ -19,23 +19,45 @@
...
+                            }), Return.VECTOR),
+                            TypeHandler(BooleanHandler(b => b.GyroOverride, (b, v) => b.GyroOverride = v), Return.BOOLEAN))
+                        , Direction.NONE),
+                        TypeHandler(NumericHandler(GetPitch, SetPitch, 10), Direction.UP),
+                        TypeHandler(NumericHandler(b => -GetPitch(b), (b, v) => SetPitch(b, -v)), Direction.DOWN),
+                        TypeHandler(NumericHandler(b => -GetYaw(b), (b, v) => SetYaw(b, -v)), Direction.LEFT),
+                        TypeHandler(NumericHandler(GetYaw, SetYaw), Direction.RIGHT),
+                        TypeHandler(NumericHandler(GetRoll, SetRoll), Direction.CLOCKWISE),
+                        TypeHandler(NumericHandler(b => -GetRoll(b), (b, v) => SetRoll(b, -v)), Direction.COUNTERCLOCKWISE));
+
```
MerlinofMines, line 45: Missing the increments

Issue #139, opened 2022-01-11
Suprising behaviour of some casts


jgersti:
This splits the behavioral changes mentioned #121 and #131 into a seperate issue.

The following cast behaviour is surprising.
- bool <-> number
    `true -> 0 -> false`
     `false -> -1 -> false`
- color <-> vector
    `255:255:255 -> #FFFFFF -> 1.0:1.0:1.0 -> #010101 -> 0.003922:0.003922:0.003922 -> #000000 -> 0:0:0`

MerlinofMines:
Resolved by #140

PR #140, opened 2022-01-11
Change cast behaviour

jgersti:
closes issue #139
depends on #131 for the presence of tests

- [x] resolve merge conflict against #131 
- [x] change test to reflect new behaviour (depends on #131)


jgersti:
SInce the previous behaviour was never documented, i have no idea what form the documentation should have.

EasyCommands/Common/Operations.cs
```diff
@@ -89,11 +89,11 @@ partial class Program {
             AddBiOperation<bool, bool>(BiOperand.COMPARE, (a, b) => a.CompareTo(b));
             AddBiOperation<string, string>(BiOperand.COMPARE, (a, b) => a.CompareTo(b));
             AddBiOperation<float, float>(BiOperand.COMPARE, (a, b) => a.CompareTo(b));
-            AddBiOperation<Vector3D, Vector3D>(BiOperand.COMPARE, (a, b) => a.Equals(b));
+            AddBiOperation<Vector3D, Vector3D>(BiOperand.COMPARE, (a, b) => !a.Equals(b));
             AddBiOperation<Color, Color>(BiOperand.COMPARE, (a, b) => a.PackedValue.CompareTo(b.PackedValue));
             AddBiOperation<Vector3D, float>(BiOperand.COMPARE, (a, b) => a.Length().CompareTo(b));
             AddBiOperation<float, Vector3D>(BiOperand.COMPARE, (a, b) => a.CompareTo(b.Length()));
-            AddBiOperation<KeyedList, KeyedList>(BiOperand.COMPARE, (a, b) => Enumerable.SequenceEqual(a.keyedValues, b.keyedValues));
+            AddBiOperation<KeyedList, KeyedList>(BiOperand.COMPARE, (a, b) => !Enumerable.SequenceEqual(a.keyedValues, b.keyedValues));
```
MerlinofMines, line 96: This doesn't seem right to me.  I don't understand how "myList is empty" continues to work with this change.

I ran tests locally after changing these and the tests failed for me.  Not sure what's going on here.  

We should add SimpleComparisonTests to test all the expected comparisons, similar to what we did for casting.  

Let me know if you don't really want to take that on and I can try to tackle it.

jgersti: `Compare` should return  `less/-1`, `equal/0` or `greater/1` and is only really welldefined for types that can be ordered, like numbers or strings (lexographic).
Neither `vector` nor `list` can be ordered, they can only  be checked for equality.
Atleast to my knowledge it is customary to return `greater/1` for  inequality in this case.

So simply negating the the equality is shorter than `expr ? 0 : 1` to restore the proper integer values, because the `BiOperation`s wrap the result into an `Primitive` that needs to be unpacked on `Compare` (via a cast to a number) anyway.
This only works because the proper `number` to `bool` (and vice versa) cast behaviour is implemented now.

jgersti: To you want  the tests in a seperate PR or in here? I can take a look at that over the weekend.

MerlinofMines: Let's go ahead and add them here just for completeness.  Thank you for tackling this one!

EasyCommands.Tests/EasyCommands.Tests.csproj
```diff
@@ -173,11 +173,16 @@
     <Compile Include="ScriptTests\FunctionalTests\ControlCommandTests.cs" />
     <Compile Include="ScriptTests\FunctionalTests\IterationCommandTests.cs" />
     <Compile Include="ScriptTests\FunctionalTests\SimpleWaitCommandTests.cs" />
-    <Compile Include="ScriptTests\FunctionalTests\SimpleComparisonTests.cs" />
+    <Compile Include="ScriptTests\FunctionalTests\SimpleComparisonsOtherTests.cs" />
```
MerlinofMines, line 176: I'm hoping we can just get rid of these tests, presuming they are covered by the new tests.

EasyCommands.Tests/EasyCommands.Tests.csproj
```diff
@@ -173,11 +173,16 @@
...
     <Compile Include="ScriptTests\FunctionalTests\SimpleAggregationTests.cs" />
     <Compile Include="ScriptTests\FunctionalTests\SimpleColorTests.cs" />
     <Compile Include="ScriptTests\FunctionalTests\SimpleVectorTests.cs" />
     <Compile Include="ScriptTests\FunctionalTests\SimpleCastTests.cs" />
+    <Compile Include="ScriptTests\FunctionalTests\SimpleComparisonsBooleanTests.cs" />
+    <Compile Include="ScriptTests\FunctionalTests\SimpleComparisonsNumberTests.cs" />
+    <Compile Include="ScriptTests\FunctionalTests\SimpleComparisonsStringTests.cs" />
+    <Compile Include="ScriptTests\FunctionalTests\SimpleComparisonsVectorTests.cs" />
+    <Compile Include="ScriptTests\FunctionalTests\SimpleComparisonsListTests.cs" />
```
MerlinofMines, line 185: Looks like we need to add Color Tests.

EasyCommands.Tests/ScriptTests/FunctionalTests/SimpleComparisonsBooleanTests.cs
```diff
@@ -0,0 +1,92 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace EasyCommands.Tests.ScriptTests {
+    [TestClass]
+    public class SimpleComparisonsBooleanTests {
+        [TestMethod]
+        public void LessThan() {
+            var lines = new List<String>{ "False < True", "True < False", "False < False", "True < True" };
+
+            using (var test = new SimpleExpressionsTest(lines)) {
```
MerlinofMines, line 12: I hate to do this to you, but I think SimpleExpressionsTest is unnecessary.  I'd prefer to follow the pattern used in all the other tests.

Instead, you can do the following with equal effect:

var script = @"
print false < true
print true < false
print false < false
print true < true
";

EasyCommands.Tests/EasyCommands.Tests.csproj
MerlinofMines, line 185: Awesome test coverage, thank you!

EasyCommands/Common/Operations.cs
```diff
@@ -89,11 +89,11 @@ partial class Program {
             AddBiOperation<bool, bool>(BiOperand.COMPARE, (a, b) => a.CompareTo(b));
             AddBiOperation<string, string>(BiOperand.COMPARE, (a, b) => a.CompareTo(b));
             AddBiOperation<float, float>(BiOperand.COMPARE, (a, b) => a.CompareTo(b));
-            AddBiOperation<Vector3D, Vector3D>(BiOperand.COMPARE, (a, b) => a.Equals(b));
+            AddBiOperation<Vector3D, Vector3D>(BiOperand.COMPARE, (a, b) => !a.Equals(b));
             AddBiOperation<Color, Color>(BiOperand.COMPARE, (a, b) => a.PackedValue.CompareTo(b.PackedValue));
             AddBiOperation<Vector3D, float>(BiOperand.COMPARE, (a, b) => a.Length().CompareTo(b));
-            AddBiOperation<float, Vector3D>(BiOperand.COMPARE, (a, b) => a.CompareTo(b.Length()));
-            AddBiOperation<KeyedList, KeyedList>(BiOperand.COMPARE, (a, b) => Enumerable.SequenceEqual(a.keyedValues, b.keyedValues));
+            AddBiOperation<float, Vector3D>(BiOperand.COMPARE, (a, b) => a.CompareTo((float)b.Length()));
```
MerlinofMines, line 95: is the float cast actually necessary? if so we likely need it on line 94 as well

EasyCommands.Tests/ScriptTests/FunctionalTests/SimpleComparisonsVectorTests.cs
```diff
@@ -0,0 +1,104 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace EasyCommands.Tests.ScriptTests {
+    [TestClass]
+    public class SimpleComparisonsVectorTests {
+        [TestMethod]
+        public void EqualTo() {
+            var lines = new List<String> {
+                "x == 1:0:0",
```
MerlinofMines, line 11: i think the tests will read a little better without using the built in "x,y,z" variables.  Can we make these explicit vectors instead?

EasyCommands.Tests/EasyCommands.Tests.csproj
jgersti: This file contains the test for the ternary and for colors

EasyCommands.Tests/EasyCommands.Tests.csproj
jgersti: the color comparison are in the "other" test since they are incomplete (only equality/inequality).
I have no idea what sense the other comparisons would make.

EasyCommands.Tests/ScriptTests/FunctionalTests/SimpleComparisonsBooleanTests.cs
jgersti: the main point is to get the proper message for the fail messages in the asserts.

EasyCommands.Tests/ScriptTests/FunctionalTests/SimpleComparisonsVectorTests.cs
jgersti: I am not sure how much the script likes comparisons with explicit left hand sides for vector, color and list.
Something like 
```
if [-1,2] contains var
    ...
```
throws an `key not found in dictionary` exception.

EasyCommands/Common/Operations.cs
jgersti: `b.Length()` returns `double` and `CompareTo(...)` has no overloads for narrowing conversions.
On line 94 is is not strictly necessary and in most cases makes no difference but if you want these to be interchangable it should be cast to `float` as well.

EasyCommands.Tests/ScriptTests/FunctionalTests/SimpleComparisonsVectorTests.cs
MerlinofMines: huh, that sounds like a bug.  I'm curious to look into this one more

EasyCommands.Tests/ScriptTests/FunctionalTests/SimpleComparisonsBooleanTests.cs
MerlinofMines: When it comes to test, I actually don't mind that the script is exactly what you're testing against, vs having a layer of separation from the actual compiled test.  The extra stuff is just the "print" statement in this case, haha.

EasyCommands.Tests/EasyCommands.Tests.csproj
MerlinofMines: I see.  I need to re-do the structure of the tests and introduce "Operations" tests, with "comparison" folder containing all of these comparison tests.  Obviously that'll be a separate effort.

EasyCommands/Common/Operations.cs
MerlinofMines: ya we should make both of them cast to float in that case.

EasyCommands.Tests/ScriptTests/FunctionalTests/SimpleComparisonsBooleanTests.cs
jgersti: Do you want it changed?

EasyCommands/Common/Operations.cs
jgersti: Both cases use `CastNumber(ResolvePrimitive(...))` now, that should be 1 resp. 3 symbols shorter than the explicit cast after minifying.

EasyCommands.Tests/ScriptTests/FunctionalTests/SimpleComparisonsBooleanTests.cs
MerlinofMines: I would like it changed, but I can handle as part of re-doing some the Operations tests.

MerlinofMines (approved):

PR #143, opened 2022-01-16
ShipController Improvements

### LINKED ISSUE #142: Cockpit Improvements
Looks like another iteration of Cockpit Improvements is needed.

First, the "rotation" property of cockpits doesn't seem to work correctly for the arrow keys (at least on my keyboard).  Values returned are from -9 to 9, vs expected -1  to 1.  If this is consistent across all keyboards then a fix should be put in place to scale these to -1, 1.

Second, there's a requset to add "is occupied" support for cockpits (and could also be cryo chambers?).  

This is definitely available on IMyShipController but isn't exposed as a property, so an explicit property mapping will be needed to expose this.

As part of this, perhaps Cryo Chambers can be added as a new block handler?

MerlinofMines:
First, this commit adds a new handler for Cryo Chambers, which extend from Cockpits.  Second, this commit adds a new "Use" property that can be used to get whether cockpits, cryo chambers, and/or remote controls are currently in use (occupied or being controlled).  Third, this commit fixes some issues with input and rotation properties to make the default input forwards and scale rotational input to -1, 1.

This commit resolves #142


docs/EasyCommands/blockHandlers/cockpit.md
```diff
@@ -152,7 +188,7 @@ if "My Cockpit" upwards velocity > 0
 * Takes an optional Direction attribute
 * Keywords: ```input, inputs, user, users, pilot, pilots```
 
-If a direction is not included, Returns a Vector (Right:Up:Backwards) representing the magnitude of movement input (values between 0 - 1) from the engineer sitting in the cockpit.
+If a direction is not included, Returns a Vector (Right:Up:Forwards) representing the magnitude of movement input (values between 0 - 1) from the engineer sitting in the cockpit.
```
MerlinofMines, line 191: Probably needs to be updated in Remote Control as well

MerlinofMines: nope, docs reference these.

Issue #122, opened 2022-01-04
FR: Unexpected behaviour of `exp` 


jgersti:
The expected behaviour of `exp` would that of the exponential function `e^x`.
If the behaviour is changed, `log` should be added as the natural logrithm.

MerlinofMines:
Id suggest on for natural log since log might also mean print

MerlinofMines:
Can you describe the issue a little bit more?

Are you saying the e^x is not working correctly at the moment?  I would expect it to, if not that's definitely a bug, but i'm pretty sure there's a test covering this.

Note that operations are dependent on the input types, and that there is only 1 operation for "exp, ^, xor" currently.  On booleans it's XOR, on numbers is a^b.

Thanks!

jgersti:
The expected meaning of `exp` is the exponential function (`exp(x) = e^x`). 
right now it has the meaning of the power operator, which is unexpected.

MerlinofMines:
I see what you mean.  Since e is already a constant and ^ an operation, I think we can just remove exp keywords

MerlinofMines:
I've gone ahead and removed "exp" keywords from the Exponent Operation.  I'm supportive of adding "ln" for natural log as a new Uni operation against numeric types, if we decide that's useful enough to add.  I think "e^x" is good enough for "exp" support.  And I think we should leave "log" for Print statements.

PR #144, opened 2022-01-17
Support nautral log as UniOperand

msetze01:
As suggested in #122 - implements natural log unioperand as "ln"


pandincus (approved):

MerlinofMines:
Nice!  one minor nit: Can we get rid of the "exp" keywords, so that we can resolve the linked issue?

Issue #158, opened 2022-01-23
Add WorldMatrix property and Matrix operations


jgersti:
The only major stumpling block to replicate most simple standalone scripts is the ability to transform coordinates and directions into usable information via the world transformation (in reference to a control block) encode in the world matrices.

The most notable example is controlling gyroscopes in reference to the controller orientation (instead of its own).

Since the proper world matrix to use is usually the one obtained from the cockpit/remote these transformation cannot really performed on the block which is manipulated.

As a first step into that direction I would propose to add a way to access and use world matrices.
Support for arbitrary matrices can be added further down the line.

---

Since the matrices are 4x4 but vector has 3 components there is no difference between position vectors (4th component would be 1 or a scaling factor) and directional vectors (4th component would be 0), there are 2 different matrix multiplications _(these are left multiplications by row vectors ?!?)_.
`Transform` is for position vectors and `TransformNormal` is for directional vectors.
Apart from the normal operations (`+`, `-`, `*`_[matrix-matrix, scalar-matrix, matrix-scalar]_, `/` _[matrix-scalar only]_),  the `Transpose`, `Invert` and `Rescale` _(this is not the same a multiplying with a number!)_ operations are also important.
These matrices also have apart from the usual access by element, the six cardinal direction and the `Translation` properties to access  the (truncated) rows.

---

All in all this is a rather comprehensive feature request that would allow the (advanced) user to replace many simple script.
For documentation the we should ask Whiplash141 if we can use [Vector Transformation with World Matrices](https://github.com/malware-dev/MDK-SE/wiki/Vector-Transformations-with-World-Matrices) or just reference there for an explanation when to use what.

MerlinofMines:
Can you describe specifically what you would expect to write as a script if you had this functionality?  Like an example script of what you could do vs what you have to do right now?

jgersti:
Right now we can do  nothing of the sort since the information is simply not available.

I would like to replace simple scripts for gyro/subGrid/steering control

The simplest example is controling gyros relative to a cockpit/remote.
```
var rotationVec = new Vector3D(pitchSpeed, yawSpeed, rollSpeed);
// transform desired rotation speeds into world coordinates
var relativeRotationVec = Vector3D.TransformNormal(rotationVec, referenceWorldMatrix);

foreach (var g in gyroList)
{
    // transform rotation speeds from world to gyro coordinate system
[...]

PR #165, opened 2022-01-24
Simplify Block Direction Code

jgersti:
This PR replaces the current code by reading the directions directly from the blocks WorldMatrix.

The old code is still there.

see #158

### DIFF
--- a/EasyCommands.Tests/ScriptTests/MockEntityUtility.cs
+++ b/EasyCommands.Tests/ScriptTests/MockEntityUtility.cs
@@ -69,6 +69,10 @@ namespace EasyCommands.Tests.ScriptTests {
             mockShip.Setup(b => b.WorldMatrix).Returns(MatrixD.Identity);
         }
 
+        public static void MockWorldMatrix<T>(Mock<T> mockBlock, Vector3D position, Vector3D forward, Vector3D up) where T : class, IMyTerminalBlock {
+            mockBlock.Setup(b => b.WorldMatrix).Returns(MatrixD.CreateWorld(position, forward, up));
+        }
+
         public static void MockOrientation<T>(Mock<T> mockBlock, Vector3 forward, Vector3 up) where T : class, IMyTerminalBlock {
             Mock<IMyCubeGrid> mockGrid = new Mock<IMyCubeGrid>();
             mockGrid.Setup(g => g.GridIntegerToWorld(new Vector3I(0, 0, 0))).Returns(new Vector3D(0, 0, 0));
--- a/EasyCommands/BlockHandlers/TerminalBlockHandlers.cs
+++ b/EasyCommands/BlockHandlers/TerminalBlockHandlers.cs
@@ -79,12 +79,12 @@ namespace IngameScript {
                 AddPropertyHandler(ValueProperty.ACTION, new SimplePropertyHandler<T>((b, p) => p.attributeValue.GetValue(), (b, p, v) => b.ApplyAction(CastString(p.attributeValue.GetValue())), ResolvePrimitive(0)));
 
                 AddDirectionHandlers(Property.DIRECTION, Direction.FORWARD,
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Forward)), Direction.FORWARD),
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Backward)), Direction.BACKWARD),
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Up)), Direction.UP),
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Down)), Direction.DOWN),
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Left)), Direction.LEFT),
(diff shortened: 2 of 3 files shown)


EasyCommands.Tests/ScriptTests/MockEntityUtility.cs
```diff
@@ -69,6 +69,10 @@ class MockEntityUtility {
             mockShip.Setup(b => b.WorldMatrix).Returns(MatrixD.Identity);
         }
 
+        public static void MockWorldMatrix<T>(Mock<T> mockBlock, Vector3D position, Vector3D forward, Vector3D up) where T : class, IMyTerminalBlock {
+            mockBlock.Setup(b => b.WorldMatrix).Returns(MatrixD.CreateWorld(position, forward, up));
+        }
+
         public static void MockOrientation<T>(Mock<T> mockBlock, Vector3 forward, Vector3 up) where T : class, IMyTerminalBlock {
             Mock<IMyCubeGrid> mockGrid = new Mock<IMyCubeGrid>();
             mockGrid.Setup(g => g.GridIntegerToWorld(new Vector3I(0, 0, 0))).Returns(new Vector3D(0, 0, 0));
```
MerlinofMines, line 78: Should this be removed now?  Or refactored to use the WorldMatrix?

EasyCommands/BlockHandlers/TerminalBlockHandlers.cs
```diff
@@ -79,12 +79,12 @@ public class TerminalBlockHandler<T> : BlockHandler<T> where T : class, IMyTermi
...
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Up)), Direction.UP),
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Down)), Direction.DOWN),
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Left)), Direction.LEFT),
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Right)), Direction.RIGHT));
+                    TypeHandler(VectorHandler(b => b.WorldMatrix.Forward), Direction.FORWARD),
+                    TypeHandler(VectorHandler(b => b.WorldMatrix.Backward), Direction.BACKWARD),
+                    TypeHandler(VectorHandler(b => b.WorldMatrix.Up), Direction.UP),
+                    TypeHandler(VectorHandler(b => b.WorldMatrix.Down), Direction.DOWN),
+                    TypeHandler(VectorHandler(b => b.WorldMatrix.Left), Direction.LEFT),
+                    TypeHandler(VectorHandler(b => b.WorldMatrix.Right), Direction.RIGHT));
```
MerlinofMines, line 87: Are these vectors still normalized?  Part of the intention was that all of the directional vectors had a magnitude of 1, to make scaling easy without having to manually normalize directions for every usage.

EasyCommands/BlockHandlers/TerminalBlockHandlers.cs
```diff
@@ -79,12 +79,12 @@ public class TerminalBlockHandler<T> : BlockHandler<T> where T : class, IMyTermi
                 AddPropertyHandler(ValueProperty.ACTION, new SimplePropertyHandler<T>((b, p) => p.attributeValue.GetValue(), (b, p, v) => b.ApplyAction(CastString(p.attributeValue.GetValue())), ResolvePrimitive(0)));
 
                 AddDirectionHandlers(Property.DIRECTION, Direction.FORWARD,
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Forward)), Direction.FORWARD),
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Backward)), Direction.BACKWARD),
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Up)), Direction.UP),
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Down)), Direction.DOWN),
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Left)), Direction.LEFT),
-                    TypeHandler(VectorHandler(b => Normalize(GetBlock2WorldTransform(b).Right)), Direction.RIGHT));
```
MerlinofMines, line 87: I'm pretty sure we can remove the GetBlock2WorldTransform(b) method completely after changing this, which would save on a decent number of characters.  Is that right?

EasyCommands/BlockHandlers/TerminalBlockHandlers.cs
jgersti: `WorldMatrix` is constructed as normalized. You can check it by supplying non normalized vectors into `Matrix.CreateWorld` and checking the directions.

EasyCommands/BlockHandlers/TerminalBlockHandlers.cs
jgersti: I think so.

EasyCommands.Tests/ScriptTests/MockEntityUtility.cs
jgersti: I added a Mock for `WorldMatrix`, so this function is rather useless right now. But there is a orientation property in the block interface that might need to be mocked in the future?

MerlinofMines (approved):

Issue #159, opened 2022-01-23
Vectors cannot be constructed from Variables


jgersti:
It would be nice if a vector could be constructed directly constructed from variables.
A workaround right now is constructing an appropriate string and casting it into a vector.

MerlinofMines:
Interesting.  Perhaps we could create a new TernaryOperand that takes <variable> : <variable> : <variable> and use that for constructing a vector.

so a:b:c would yield Vector(a,b,c).

jgersti:
The tokenizing step is doing the right thing already.
If a variable is contained in the vector expression the parser resolve this to:
```
... [variable] ':' [variable] ':' [variable] ...
```
After that the parser gets confused because there are no rules for resolving this yet.
Something akin to the handling of lists should be feasible.

On a site note: Can you create a list that references a variable/expression?

MerlinofMines:
Yep, that matches my expectation.  A new processing rule is needed to match the above to a new "DynamicVariable" that returns a vector using the expression.

Indeed you can create a list that references a variable / expression.  The trick is to bind the list, vs setting it (similar to other bound variables
For example:

```
set a to 1
set b to 2
set c to 3
bind myList to [a,b,c]
print myList
#[1,2,3]

set a to 4
print myList
#[4,2,3]
```

This isn't clear from Bound Variables docs, or collections docs.  Could be added.

It would work similarly for dynamic vectors I believe.

```
set a to 1
set b to 2
set c to 3

set myVector to a:b:c
bind myBoundVector to a:b:c

set a to 4

print myVector
#1:2:3

print myBoundVector
#4:2:3




```

MerlinofMines:
This could have some overlap with #151 

I wonder if we can/should add another primitive type for a GPS coordinate, with a parsing rule for name:a:b:c:color = GPS coordinate.  GPS coordinate would have a label, a vector position, and a color.  You could cast a GPS to a vector, or to a color.  String would print out the GPS coordinate directly.  Getting the label might require the use of string splitting, not sure, since CastString(gpsCoordinate) should print out the full GPS coordinate, not just the label.

PR #166, opened 2022-01-25
Refactor `ParseSeperatorTokens`

jgersti:
While investigating #159 i found some unneeded code.
Instead of first splitting, joining, copying (or rather: constructing a new string from the old ones char array that is cast to a list) and then splitting again, the the first split operation is returned directly. Also the loops to add spaces around the tokens got moved into its own method.

For a few characters more `ParseSurrounded`, `ParseSeperateTokens` and the lambda in `ParseTokens` could return `IEnumerator<...>` and the `toArray` calls could be removed.
This would make these functions truely lazy and maybe give a small performance boost on the tokenizing step.

### DIFF
--- a/EasyCommands/CommandParsers/ParameterParsers.cs
+++ b/EasyCommands/CommandParsers/ParameterParsers.cs
@@ -464,19 +464,19 @@ namespace IngameScript {
                     : new Token[] { new Token(element, true, characters.Length > 1) })  // Keep the entire item
                 .ToArray();
 
-        String[] ParseSeparateTokens(String command) {
-            var newCommand = command;
-            foreach (var token in separateTokensFirstPass) newCommand = newCommand.Replace(token, " " + token + " ");
-            newCommand = string.Join(" ", newCommand.Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries).SelectMany(token => {
+        String[] ParseSeparateTokens(String command) =>
+            AddSpaceArroundTokens(command, separateTokensFirstPass)
+            .Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries)
+            .SelectMany(token => {
                 Primitive ignored;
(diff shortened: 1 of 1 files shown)


EasyCommands/CommandParsers/ParameterParsers.cs
```diff
@@ -464,19 +464,19 @@ partial class Program {
...
+            .Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries)
+            .SelectMany(token => {
                 Primitive ignored;
                 if (separateTokensFirstPass.Contains(token) || ParsePrimitive(token, out ignored)) return new[] { token };
-                foreach (var s in separateTokensSecondPass) token = token.Replace(s, " " + s + " ");
-                return token.Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
-            }));
-
-            List<char> commandArray = newCommand.ToCharArray().ToList();
+                return AddSpaceArroundTokens(token, separateTokensSecondPass).Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
```
MerlinofMines, line 473: Nice!  nit: can we use the ternary operator to remove the two "return" statements?

return (separateTokensFirstPass.Contains(token) || ParsePrimitive(token, out ignored)) ? new[] { token} : AddSpaceAroundTokens(....

?

EasyCommands/CommandParsers/ParameterParsers.cs
```diff
@@ -464,19 +464,19 @@ partial class Program {
...
+            }).ToArray();
 
-            return new string(commandArray.ToArray()).Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
+        String AddSpaceArroundTokens(String command, String[] tokens) {
```
MerlinofMines, line 476: nit: "AddSpaceAroundTokens" (typo)

MerlinofMines (approved):

PR #169, opened 2022-01-25
Variable Vectors

jgersti:
This is a quick and dirty attempt to implement vectors with variable components, see #159 

There are several issues with this implementation.
* there are priority issues between the operators and the vector declaration
* directly parsing strings had to be disabled to resolve issues between the ternary operator and the vector declaration. As a side effect all test that have quotes around the vector to not work.
* gps parsing in its current form broke

the first point can as far as i can see not be resolved without having delimiters around the vector.

### DIFF
--- a/EasyCommands.Tests/ScriptTests/FunctionalTests/Variables/VectorVariableTests.cs
+++ b/EasyCommands.Tests/ScriptTests/FunctionalTests/Variables/VectorVariableTests.cs
@@ -52,5 +52,32 @@ Print false?a:0:0:1:0:0
                 Assert.AreEqual("1:0:0", test.Logger[1]);
             }
         }
+
+        [TestMethod]
+        public void AssignUniOperandVariableVector() {
+            var script = @"
+assign vec to sqrt 4:round 0.5:-2
+Print vec
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("2:0:-4", test.Logger[0]);
+            }
+        }
+
+        [TestMethod]
+        public void AssignBiOperandVariableVector() {
+            var script = @"
+assign a to 1
+assign vec to a:0:0 + a:0:0 + a + 1
+Print vec
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("4:0:0", test.Logger[0]);
+            }
+        }
     }
 }
--- a/EasyCommands.Tests/ScriptTests/FunctionalTests/Variables/VectorVariableTests.cs
+++ b/EasyCommands.Tests/ScriptTests/FunctionalTests/Variables/VectorVariableTests.cs
@@ -62,7 +62,7 @@ Print vec
             using (var test = new ScriptTest(script)) {
                 test.RunOnce();
 
-                Assert.AreEqual("2:0:-4", test.Logger[0]);
+                Assert.AreEqual("2:0:-2", test.Logger[0]);
             }
         }
 
--- a/EasyCommands.Tests/ScriptTests/FunctionalTests/Variables/VectorVariableTests.cs
+++ b/EasyCommands.Tests/ScriptTests/FunctionalTests/Variables/VectorVariableTests.cs
@@ -5,7 +5,7 @@ namespace EasyCommands.Tests.ScriptTests {
     [TestClass]
     public class VectorVariableTests {
         [TestMethod]
-        public void AssignVariableVector() {
+        public void Assign() {
             var script = @"
 assign a to 1
 bind b to a + 1
@@ -20,7 +20,7 @@ Print vec
         }
 
         [TestMethod]
-        public void BindVariableVector() {
+        public void Bind() {
             var script = @"
 assign a to 1
 bind b to a + 1
@@ -38,23 +38,7 @@ Print vec
         }
 
         [TestMethod]
-        public void TernaryAndVariableVector() {
-            // worst case formatting
-            var script = @"
-assign a to 2
-Print true?a:0:0:1:0:0
-Print false?a:0:0:1:0:0
-";
-            using (var test = new ScriptTest(script)) {
-                test.RunOnce();
-
-                Assert.AreEqual("2:0:0", test.Logger[0]);
-                Assert.AreEqual("1:0:0", test.Logger[1]);
-            }
-        }
-
-        [TestMethod]
-        public void AssignUniOperandVariableVector() {
+        public void AssignUniOperands() {
             var script = @"
 assign vec to sqrt 4:round 0.5:-2
 Print vec
@@ -67,7 +51,7 @@ Print vec
         }
 
         [TestMethod]
-        public void AssignBiOperandVariableVector() {
+        public void AssignBiOperands() {
             var script = @"
 assign a to 1
 assign vec to a:0:0 + a:0:0 + a + 1
@@ -77,6 +61,130 @@ Print vec
                 test.RunOnce();
 
                 Assert.AreEqual("4:0:0", test.Logger[0]);
+            }
+        }
+
+        [TestMethod]
+        public void TernaryLeft() {
+            // worst case formatting
+            var script = @"
+assign a to 2
+Print true ? a:0:0 : a
+Print false ? a:0:0 : a
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("2:0:0", test.Logger[0]);
+                Assert.AreEqual("2", test.Logger[1]);
+            }
+        }
+
+        [TestMethod]
+        public void TernaryRight() {
+            // worst case formatting
+            var script = @"
+assign a to 2
+Print true ? a : 0:0:a
+Print false ? a : 0:0:a
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("2", test.Logger[0]);
+                Assert.AreEqual("0:0:2", test.Logger[1]);
+            }
+        }
+
+        [TestMethod]
+        public void TernaryBoth() {
+            // worst case formatting
+            var script = @"
+assign a to 2
+Print true ? a:0:0 : 0:0:a
+Print false ? a:0:0 : 0:0:a
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("2:0:0", test.Logger[0]);
+                Assert.AreEqual("0:0:2", test.Logger[1]);
+            }
+        }
+
+        [TestMethod]
+        public void TernaryBothPrimitiveLeft() {
+            // worst case formatting
+            var script = @"
+assign a to 2
+Print true ? 1:2:3 : 0:0:a
+Print false ? 1:2:3 : 0:0:a
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("1:2:3", test.Logger[0]);
+                Assert.AreEqual("0:0:2", test.Logger[1]);
+            }
+        }
+
+        [TestMethod]
+        public void TernaryBothPrimitiveRight() {
+            // worst case formatting
+            var script = @"
+assign a to 2
+Print true ? a:0:0 : 1:2:3
+Print false ? a:0:0 : 1:2:3
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("2:0:0", test.Logger[0]);
+                Assert.AreEqual("1:2:3", test.Logger[1]);
+            }
+        }
+
+        [TestMethod]
+        public void TernaryBothPrimitiveBoth() {
+            // worst case formatting
+            var script = @"
+Print true ? 3:2:1 : 1:2:3
+Print false ? 3:2:1 : 1:2:3
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("3:2:1", test.Logger[0]);
+                Assert.AreEqual("1:2:3", test.Logger[1]);
+            }
+        }
+
+        [TestMethod]
+        public void TernaryBothNoSpace() {
+            var script = @"
+assign a to 2
+Print true?a:0:0:0:a:0
+Print false?a:0:0:0:a:0
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("2:0:0", test.Logger[0]);
+                Assert.AreEqual("0:2:0", test.Logger[1]);
+            }
+        }
+
+        [TestMethod]
+        public void NestedExpressions() {
+            var script = @"
+assign a to 2
+assign vec to ((a ^ a) % a):(0:a:0 as number):(red . 1:0:0)
+Print vec
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("0:2:255", test.Logger[0]);
             }
         }
     }
--- a/EasyCommands.Tests/ScriptTests/FunctionalTests/Variables/VectorVariableTests.cs
+++ b/EasyCommands.Tests/ScriptTests/FunctionalTests/Variables/VectorVariableTests.cs
@@ -1,0 +1,56 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace EasyCommands.Tests.ScriptTests {
+    [TestClass]
+    public class VectorVariableTests {
+        [TestMethod]
+        public void AssignVariableVector() {
+            var script = @"
+assign a to 1
+bind b to a + 1
+assign vec to a:0:b
+Print vec
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("1:0:2", test.Logger[0]);
+            }
+        }
+
+        [TestMethod]
+        public void BindVariableVector() {
+            var script = @"
+assign a to 1
+bind b to a + 1
+bind vec to a:0:b
+Print vec
+a+=1
+Print vec
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("1:0:2", test.Logger[0]);
+                Assert.AreEqual("2:0:3", test.Logger[1]);
+            }
+        }
+
+        [TestMethod]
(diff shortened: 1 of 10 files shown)


jgersti:
I reenabled the parsing of _primitive_ vectors and made sure that `PrimitiveProcessor`  enclosed the `StaticVariable` in an `VectorVariable`in case a the primitive is a vector.

I also moved the vector parsing rule after the unitary operation rules so that expressions like
```
set a to 42
set vec to -a:0:1
```
do the expected thing and yield `-42:0:1` instead of `-42:0:-1`. This behaviour should be acceptable.

The unit test fails because of a critical bug in the master branch.

EasyCommands/CommandParsers/ParameterProcessors.cs
```diff
@@ -136,7 +136,11 @@ public class PrimitiveProcessor<T> : ParameterProcessor<T> where T : class, Prim
 
                 Variable variable = new AmbiguousStringVariable(value.value);
 
-                if (primitive != null || value.isExplicit) variable = new StaticVariable(primitive ?? ResolvePrimitive(value.value));
+                if (primitive != null || value.isExplicit) {
+                    variable = new StaticVariable(primitive ?? ResolvePrimitive(value.value));
+                    if (variable.GetValue().returnType == Return.VECTOR)
+                        variable = new VectorVariable(variable);
+                }
```
MerlinofMines, line 143: I don't think we need to adjust this one.  This is only in place because we are trying to use VectorVariables for all instances of using Vector, and I don't think that's necessary.

I think we de-scope "VectorVariable" to only be used if you are creating a Vector via variable:variable:variable.

The case of float:float:float can be left alone.  It's ok that vectors of this form aren't processed as primitives but rather get processed as a dynamic type of variable later by the parsing rule.

Note this also great simplifies the code needed in VectorVariable

MerlinofMines: This probably makes GPS parsing challenging, as you likely found out and what also may have led you to try and combine these.  

Perhaps the answer is that we stop trying to parse vectors at all inside of ParsePrimitive, insteaad relying only on Processing Rules.  Floating point numbers will still get resolved to static variables, at which point they will then get parsed into either a GPS variable (var:var:var:var:var), otherwise a vector (var:var:var), otherwise possibly a ternary (var?var:var)

EasyCommands/CommandParsers/ProcessingEngine.cs
```diff
@@ -154,6 +154,11 @@ partial class Program {
             OneValueRule(Type<UniOperationCommandParameter>, requiredRight<VariableCommandParameter>(),
                 (p, df) => new VariableCommandParameter(new UniOperandVariable(p.value, df.value))),
 
+            //DynamicVectorProcessor
+            FourValueRule(Type<TernaryConditionSeparatorParameter>, requiredLeft<VariableCommandParameter>(), requiredRight<VariableCommandParameter>(), requiredRight<TernaryConditionSeparatorParameter>(), requiredRight<VariableCommandParameter>(),
```
MerlinofMines, line 158: Seems like we'll want to change the name to ColonCommandParameter

EasyCommands/CommandParsers/ProcessingEngine.cs
```diff
@@ -154,6 +154,11 @@ partial class Program {
...
+                (sep1, x, y, sep2, z) => AllSatisfied(x, y, z) && !(x.GetValue().value is VectorVariable || y.GetValue().value is VectorVariable || z.GetValue().value is VectorVariable),
```
MerlinofMines, line 159: You can't use x.Getvalue().value here, as during parsing you don't actually know the value of the variable.

I think you simply look for the pattern var:var:var and treat that as a dynamic VectorVariable.

Only at runtime can you actually determine.  In that code I wouldn't do too much.  A simple CastNumber(v) should be sufficient.  Perhaps we can wrap it with a try catch to throw a clear "invalid variable used in vector creation" message or similar.

EasyCommands/Common/Variables.cs
```diff
@@ -55,6 +55,34 @@ public class TernaryConditionVariable : Variable {
...
+                        return vec.GetValue();
+                    else
+                        throw new Exception("Invalid Vector variable.");
+                }
+                else {
+                    if (X.GetValue().returnType == Return.NUMERIC && Y.GetValue().returnType == Return.NUMERIC && Z.GetValue().returnType == Return.NUMERIC)
+                        return ResolvePrimitive(new Vector3D(CastNumber(X.GetValue()), CastNumber(Y.GetValue()), CastNumber(Z.GetValue())));
+                    else
+                        throw new Exception("Invalid Vector variable.");
+                }
```
MerlinofMines, line 82: Let's de-scope VectorVariable to only handle the case of dynamically creating a Vector via a:b:c.

In the VectorVariable code I wouldn't do too much for checking the return types are numeric..  A simple CastNumber(v) should be sufficient.  Perhaps we can wrap it with a try catch to throw a clear "invalid variable used in vector creation" message or similar.

EasyCommands/CommandParsers/ProcessingEngine.cs
jgersti: This only resolves down to `Variable` not to its primitive value.
That pattern is repeatedly used to check the actual type of the `Variable`.

EasyCommands/CommandParsers/ParameterProcessors.cs
jgersti: * You can try the first commit to see what happens if you remove the parsing of _primitive_ vectors. Several of the test fail because they rely either on the direct parsing in the tokenizing step or on line 140 above. Also #170 is in full effect then!

* Unfortunatly i found that i had to wrap every vector into a `VectorVariable` so i can distinguish `VectorVariable`s from other Variables. This is necessary otherwise the vector rule is to greedy, see discussion below.

* The snippet above only affects _primitive_ vectors already, all other cases are handled by the new vector rule.

* What form should that even take? I still can not understand why `ParseVectorFromGPSCoordinate` even works right now.

* First see the first point. Secondly `var ? var:var:var : var:var:var`, `var? var : var:var:var:var:var` and `var? var:var:var:var:var : var` cannot be distinguished __even if__ we find a way to define vector and gps rules that not so greedy to consume themself and eachother again (see discusion below).

EasyCommands/Common/Variables.cs
jgersti: `CastNumber is too lenient. If the user manages to stuff a vector into one component (which is possible with a few tricks) that gets resolved to its length and the evaluated vector would appaer valid.

EasyCommands/CommandParsers/ParameterProcessors.cs
MerlinofMines: This isn't the first instance of a command line being potentially ambiguous, and it won't be the last unfortunately.

I think we can make it clear in the docs that  ```var?var:var:var``` is always parsed as var ? (var:var:var) by default.  A user can clarify their intent, if needed, using parentheses: ```(var ? var : var) : var : var```, or ```var ? (var:var:var) : var```.  

We could change the vector dilimeter, true, though I definitely prefer it, especially given that it is the delimiter used by Space Engineers for GPS coords.  I think we want to keep the ability to directly parse a GPS coordinate for most cases.

I think we fix the critical parsing bug first, and then remove vector parsing entirely from Primitive Parsing, relying on the rules instead.  And we make it clear in the docs on both Ternary Operation and Vector Creation that if you use the two together the default behavior is used and to use parentheses if needed.

I apologize, I haven't had any time to sit down and look at some of this in detail; I should have some time tomorrow night to really look at this.

MerlinofMines: Given the default behavior, I don't think the check against VectorVariable would be needed anymore.  The rule would simply be "if you see var : var : var, that's a vector.  And since we parse lines from right to left, the behavior will be consistent.

EasyCommands/Common/Variables.cs
MerlinofMines: Fair enough I suppose.  I'm debating whether that's a bug or a feature.  

You can simplify I think:

if(NewList(x,y,z).TrueForAll(x => x.GetValue().returnType == Return.Numeric) return ResolvePrimitive( .... )
throw new Exception("Invalid Vector Variables");

EasyCommands/CommandParsers/ProcessingEngine.cs
jgersti: The problem is that is not parsed that way without the constrained.

We have the rules
```
vector:  var : var : var -> var
ternary: var ? var : var -> var
```

Let us consider there was no constraint on the vector rule.
If we try to parse `var? var:var:var : var:var:var` with these two rules we get the follwoing sequence (token | rule, cursor) where indices are added to the variables to distinguish them.
```
v1 ? v2 : v3 : v4 : v5 : v6 : v7 | vector
                       ^
v1 ? v2 : v3 : v4 : v8 | vector
             ^
v1 ? v2 : v9 | ternary
   ^
v10
```
`v9` is an invalid vector since its z component is `v8`, another  vector.

With the constraint that the vector rule cannot consume vector variables we get the desired sequence
```
v1 ? v2 : v3 : v4 : v5 : v6 : v7 | vector'
                       ^
v1 ? v2 : v3 : v4 : v8 | vector' 
         ^
v1 ? v9 : v8 | ternary
   ^
v10
```

jgersti: GPS would be even worse since i dont see a way to constrain the rule so it does not get applied when we actually mean vectors or we are in a similar situation as above or even worse are in situation where the vector rule tries to consume a gps variable.

EasyCommands/CommandParsers/ParameterProcessors.cs
jgersti: See below for discussion on the parsing rule.

Right now several tests rely on `PrimitiveProcessor` to strip double quotes, one test checks that it actually is an primitive and then there is that gps test where i dont understand why it even works.

EasyCommands/Common/Variables.cs
jgersti: I would call it a bug.

Let me count the chars.
It is really difficult to fight my instinct to avoid allocations. 😄

jgersti:
rebased onto master

EasyCommands/CommandParsers/ProcessingEngine.cs
MerlinofMines: I have a crazy idea.

I propose that we introduce "::" as another TernaryOperationSeparator indicator, added to first pass tokens.

Then, during string parsing, we look for " : " (specifically looking for spaces) and convert to " :: ".

Then, the rules only look for AmbiguousColonSeparator for processing vectors or gps coordinates.

Vectors or gps coordinates should never be defined with spaces (we can even mention this). a : b : c  would not be processed as a vector, ever.  You must specify a:b:c (not unreasonable).

After the GPS/Vector parsing rules, we add a simple rule that converts AmbiguousColonSeparator to TernaryOperationSeperator

This approach, I believe would yield correct results, provided the user spaces correctly.

set a to b ? c:d:e : a:0:0:0:0 would be processed correctly with this approach.  The only ambiguous case would be if the user tries set a to b?c:d:e:a:0:0:0:0, which I don't think will happen because it's pretty obvious, even to the user, that it's not clear what they were trying to do.

This approach gets rid of the need for VectorVariable checking in the GPS and Vector Processing rules as well.  You can always do GPS first, and Vector second.  It also means that a user can further clarify the ternary operation intent by explicitly specifying "::" instead of ":".  We could also consider ";".

Thoughts?

MerlinofMines: I just sent out a PR for adding that behavior, which can be pushed as is if you want to rebase.  This should get you the separation desired so you can do some simple vector & gps parsing rules.

https://github.com/MerlinofMines/EasyCommands/pull/173/files

jgersti: I do not think this a good idea. It enables the user again to indiscriminately but other vectors variables into a vectors.
In the current state they have to atleast hide the vector behind an operation.

I put a few thoughts into the GPS issue. A GPS marker is a __string__ of the form `GPS:[name]:[X]:[Y]:[Z]:[color]`. 
_Note the leading `GPS`_. This can be unambigously matched with a 10 value rule or 3 rules to partially build it up.

__But__ there is actually no point to pretend that it is anything but a string since the block interface wants a string anyway. Casting such a string to vector already works (for whatever reason) and there is no use case for knowing the name or the color of a GPS marker, so a cast in the other direction is pointless. Then only some handling for properties that want a GPS (string) is needed to allow vector input and convert it into the desired format.

jgersti:
rebased onto master (840709c)

EasyCommands.Tests/ScriptTests/FunctionalTests/Variables/VectorVariableTests.cs
```diff
@@ -0,0 +1,83 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace EasyCommands.Tests.ScriptTests {
+    [TestClass]
+    public class VectorVariableTests {
```
MerlinofMines, line 6: Let's add more tests in the TernaryOperand tests testing ternary operations and vector variables combined.

Cases like ```a ? b:c:d : e``` and ```a ? b : c:d:e```

I think that'll help show why the other pull request is warranted.

We don't have tests here that actually exercise the VectorVariable check, that I can see.  We should add these to make sure the check works as expected.

EasyCommands/CommandParsers/ProcessingEngine.cs
MerlinofMines: I think we may actually need both.

As is the VectorVariable alone does not disambiguate ```a ? b : c:d:e``` from ```a ? b:c:d : e```.  For that, my pull request is needed.

The VectorVariable check helps with ```a?b:c:d:e:f:g```, interpreted as a ? b:c:d e:f:g.  I'm not even sure how reasonable it is to expect b:c:d:e:f:g to be interpreted correctly, TBD.  Most likely, a user will add spaces to clarify, and I think the parsing should use that information to help parse meaning.

I'm happy to keep the VectorVariable check as a way to protect users against themself, but I'm more concerned about handling cases where to the user it's clear what they're trying to do but it's ambiguous to the script parsing because we lost context on association via spacing.

I also don't think VectorVariable needs to be applied to

**GPS Parsing**

We can certainly solve GPS parsing separately, let's separate it from this pull request and focus on dynamic vectors instead.

EasyCommands/CommandParsers/ParameterProcessors.cs
jgersti: lets continue the discussion threads in the main thread below

EasyCommands/CommandParsers/ProcessingEngine.cs
jgersti: lets continue the discussion threads in the main thread below

EasyCommands.Tests/ScriptTests/FunctionalTests/Variables/VectorVariableTests.cs
jgersti: i have a test in a stash which i used for checking the parser steps which does cover a few more cases.
but most of the the operator behavior is covered by the old vector tests if we decide to remove primitive vectors.

jgersti:
To continue all the discussion from above in single thread:

### Primitive Vectors
If we decide to completly remove primitve vectors (`ParsePrimitive` has no vector handling anymore) the snippet in the first discussion can be reverted. But every test that uses quoted vectors and the test that cast a gps string will fail and needs to be rewritten. 
Another side effect is that all unitary operators now need parentheses to apply to all vector because they need have higher priority so vectors can have a negative first component.  With primitive vectors that was not an issue beforehand.
This behavior affects a few more tests.

_This does not really belong here, but:_ 
Additionaly removing `PrimitivePass` from the first tokenizing pass has the side effect of removing all negative number primitives!

I will write replies to the other points a bit later.

MerlinofMines:
**Vector Parsing in ParsePrimitives*
Vector handling support in parse primitives can stay, I think.   I don't think the cast to VectorVariable is strictly necessary but also isn't necessarily doing any harm.

*Removing PrimitivePass* 
Losing negative number support after removing that seems odd to me; are the first & second pass tokens set correctly?  Again, not necessarily doing harm so if still needed can stay.  Could be looked at separately.

So both of the above can likely be left as is :).

jgersti:
The second pass splits on minus.

jgersti:
### VectorVariable
The purpose of `VectorVariable` is also to make as many illformed expressions parsing errors instead of just throwing an error when the line is executed. The conversion in the primitive wrapper makes sure that primitive vectors are picked up on. 
I need to add some more test to check if your proposal messes with some of usage i envision.

I think there is in general gap in the test coverage that obvious malformed expression to not get through the parser.

jgersti:
### GPS Parsing
Some form of parsing is already supported, most likely through the bug/feature you already mentioned.
I agree that it is a seperate issue and since a valid gps string has 5 colons instead of 4 (and a fixed first token) the overlap with vector + ternary is gone anyway.

### Tests 
As said above i will add some more tests. Maybe we should also think about adding tests that the parser detects expression that are clearly meant to be invalid.

Is there a way to just invoke the tokenizing and parsing steps without trying to execute the script?

jgersti:
The tests should now cover everything that should work.  Maybe we need some coverage for expressions that should not work.

EasyCommands.Tests/ScriptTests/FunctionalTests/Variables/VectorVariableTests.cs
```diff
@@ -0,0 +1,191 @@
...
+        [TestMethod]
+        public void Assign() {
+            var script = @"
+assign a to 1
+bind b to a + 1
+assign vec to a:0:b
+Print vec
```
MerlinofMines, line 13: we should ad a+=1 and make sure that vec remains unchanged

EasyCommands.Tests/ScriptTests/FunctionalTests/Variables/VectorVariableTests.cs
```diff
@@ -0,0 +1,191 @@
...
+bind vec to a:0:b
+Print vec
+a+=1
+Print vec
+";
+            using (var test = new ScriptTest(script)) {
+                test.RunOnce();
+
+                Assert.AreEqual("1:0:2", test.Logger[0]);
+                Assert.AreEqual("2:0:3", test.Logger[1]);
```
MerlinofMines, line 36: that's pretty cool :)

MerlinofMines (approved):

Issue #170, opened 2022-01-25
Critical Bug when Parsing Floating Point Numbers


jgersti:
When tokenizing a floating point number which is directly preceded or followed by an character from `separateTokensSecondPass` the floating point number is decomposed.
`Print 2.0+1` is tokenized as `Print "2" Dot "0" Add "1"`.

In #169 this situation is even worse since the component delimiter `:` is tokenized in the second pass. This means a vector containing variables and floating point numbers cannot be defined and the dot product between two such vectors inside an the success branch of an ternary condition might be ambiguous.
`Print pi:0.5:1`is tokenized as `Print "pi" : "0" Dot "5" : "1"`

MerlinofMines:
Indeed, this one is a bit of a doozy looks like.

Let me give this one some thought.  This sort of hints at the 2 pass approach being insufficient.

Herhaps we need to introduce parsing rules, with priority ordering, similar to how the parameter processor & engine work.  Doesn't need to be near as complicated though.

Perhaps a delegate Processor taking List<Token> and returning List<Token>.

First phase pulls out firstPassTokens (renamed)
Second phase can look for floating point numbers
Third phase can parse primitives
fourth phase can parse secondPassTokens (renamed)

That approach would let us add more parsing rules iteratively as needed.  Hopefully can be done without too many characters.
Since only one line is parsed per tick (and is processed in a different tick) i don't think the additional steps for processing the tokens will cause script parsing complexity issues, but guess we'll see.

Thoughts?

jgersti:
My suggestion to keep changes pretty minimal is to split the tokenizing into the following steps:
1. _tokenize quoted substrings_ 
2. _seperate `firstPassTokens` and tokenize matches_
3. _match and tokenize primitives_ (last chance to match primitive vectors)
4. _seperate `<, >, =, &, |, ?, :` and tokenize matches_
5. _match and tokenize primitives_
6. _seperate on `+, -, .`  and tokenize matches_
7. _tokenize the reset_

Currently the following steps are performed:
1. _tokenize quoted substrings_ 
2. _seperate `firstPassTokens` and tokenize matches_
3. _match and tokenize primitives_
4. _seperate `secondPassTokens` and tokenize matches_
5. _tokenize the reset_

_As a site note:_ maybe call these functions `TokenizeXXX` since that is what they actually do.

MerlinofMines:
Ya I agree the . would need to be separated last, after checking for primitives.

[...]

PR #171, opened 2022-01-26
Rework tokenizing

jgersti:
This PR implements the changes discussed in #170 and renames the tests accordingly.

Switching from `String[]` to  `IEnumerable<String>` removes quite a number of allocations and does speed up the unit testing by about a third on my system.

_I dont like the implementation very much but C# especially v6 is limiting._

### DIFF
--- a/EasyCommands/CommandParsers/ParameterParsers.cs
+++ b/EasyCommands/CommandParsers/ParameterParsers.cs
@@ -22,8 +22,9 @@ namespace IngameScript {
         //Internal (Don't touch!)
         Dictionary<String, List<CommandParameter>> propertyWords = NewDictionary<string, List<CommandParameter>>();
 
-        string[] separateTokensFirstPass = new[] { "(", ")", "[", "]", ",", "*", "/", "!", "^", "..", "%", ">=", "<=", "==", "&&", "||", "@", "$", "->", "++", "+=", "--", "-=" };
-        string[] separateTokensSecondPass = new[] { "<", ">", "=", "&", "|", ".", "-", "+", "?", ":" };
+        string[] firstPassTokens = new[] { "(", ")", "[", "]", ",", "*", "/", "!", "^", "..", "%", ">=", "<=", "==", "&&", "||", "@", "$", "->", "++", "+=", "--", "-=" };
+        string[] secondPassTokens = new[] { "<", ">", "=", "&", "|", "-", "+", "?", ":" };
+        string[] thirdPassTokens = new[] { "." };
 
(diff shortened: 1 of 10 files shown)


EasyCommands/CommandParsers/ParameterParsers.cs
```diff
@@ -449,36 +450,41 @@ partial class Program {
...
+            Pass thirdPass = v => SeperatorPass(v, thirdPassTokens);
+            Pass secondPass = v => SeperatorPass(v, secondPassTokens, w => PrimitivePass(w, thirdPass));
+            Pass firstPass = v => SeperatorPass(v, firstPassTokens, w => PrimitivePass(w, secondPass));
+
+            return (String.IsNullOrWhiteSpace(commandString) || commandString.Trim().StartsWith("#"))
+            ? NewList<Token>()
+            : TokenizeEnclosed(commandString, "`\'\"",
                 u => u.Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries)
-                    .SelectMany(ParseSeparateTokens)
+                    .SelectMany(v => firstPass(v))
```
jgersti, line 463: The chaining of all the passes can be done as a (very long) oneliner directly inside the `SelectMany(...)` but that is unreadable.

MerlinofMines: Definitely adds a few chars, but the readability is nice.  Can always revisit later

EasyCommands/CommandParsers/ParameterParsers.cs
```diff
@@ -449,36 +450,41 @@ partial class Program {
             return commandParameters;
         }
 
-        public List<Token> ParseTokens(String commandString) => (String.IsNullOrWhiteSpace(commandString) || commandString.Trim().StartsWith("#")) ? NewList<Token>() :
-            ParseSurrounded(commandString, "`\'\"",
+        delegate IEnumerable<String> Pass(String s);
+        public List<Token> Tokenize(String commandString) {
+            Pass thirdPass = v => SeperatorPass(v, thirdPassTokens);
+            Pass secondPass = v => SeperatorPass(v, secondPassTokens, w => PrimitivePass(w, thirdPass));
+            Pass firstPass = v => SeperatorPass(v, firstPassTokens, w => PrimitivePass(w, secondPass));
```
MerlinofMines, line 457: Perhaps we can removes the double primitive pass once we get the vector rule in place :)

MerlinofMines (approved):

PR #173, opened 2022-01-27
Added ColonCommandParameter

MerlinofMines:
POC for adding a new colon command parameter.

Doesn't have tests, but shows we can use this to distinguish "a:b:c : d:e:f"

This can be done, in combination with the vector variable parsing, to enable intended behavior that works with ternary conditions, I believe, using spacing to indicate the intended behavior.

### DIFF
--- a/EasyCommands/CommandParsers/ParameterParsers.cs
+++ b/EasyCommands/CommandParsers/ParameterParsers.cs
@@ -22,7 +22,7 @@ namespace IngameScript {
         //Internal (Don't touch!)
         Dictionary<String, List<CommandParameter>> propertyWords = NewDictionary<string, List<CommandParameter>>();
 
-        string[] firstPassTokens = new[] { "(", ")", "[", "]", ",", "*", "/", "!", "^", "..", "%", ">=", "<=", "==", "&&", "||", "@", "$", "->", "++", "+=", "--", "-=" };
+        string[] firstPassTokens = new[] { "(", ")", "[", "]", ",", "*", "/", "!", "^", "..", "%", ">=", "<=", "==", "&&", "||", "@", "$", "->", "++", "+=", "--", "-=", "::"};
         string[] secondPassTokens = new[] { "<", ">", "=", "&", "|", "-", "+", "?", ":" };
         string[] thirdPassTokens = new[] { "." };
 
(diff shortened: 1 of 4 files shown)


EasyCommands/CommandParsers/ProcessingEngine.cs
```diff
@@ -237,6 +237,9 @@ partial class Program {
             FourValueRule(Type<TransferCommandParameter>, requiredRight<SelectorCommandParameter>(), requiredRight<SelectorCommandParameter>(), requiredRight<VariableCommandParameter>(), optionalRight<VariableCommandParameter>(),
                 (t, s1, s2, v1, v2) => new CommandReferenceParameter(new TransferItemCommand(s1.value, s2.value, v1.value, v2.HasValue() ? v2.GetValue().value : null))),
```
MerlinofMines, line 239: GPS & Vector Parsing Rules would need to be above this

EasyCommands/CommandParsers/ParameterParsers.cs
```diff
@@ -460,7 +462,7 @@ partial class Program {
             ? NewList<Token>()
             : TokenizeEnclosed(commandString, "`\'\"",
                 u => u.Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries)
-                    .SelectMany(v => firstPass(v))
+                    .SelectMany(v => firstPass(v.Replace(" : ", " :: ")))
```
MerlinofMines, line 465: Could be separated as it's own "Pass" but this would be the minimum needed

jgersti: I am stepping through the tokenizer and parser right now and this does not do anything right now, which defeats its prurpose.

jgersti: OK. Actually for obvious reasons.
The line above the changes removes all whitespace ...

MerlinofMines: Ya, i put it in the wrong space.  Finally have some proper time to send this out with proper tests.  Sorry, been rushing things cuz i haven't had any time.

I'll send out and update, with more tests, today

jgersti: i cherrypicked this on top of the variable vector branch and will push that in a few minutes.

PR #180, opened 2022-01-28
Fixed issue with broadcast listeners remaining active after re-parsin…

### LINKED ISSUE #167: Active Messages in Listening Queue Causes Script to halt when re-parsed
There's a but with the program execution wherein the program is not instantiated if on the first execution of the script there is an active message to process.  The script only processes the active message and then doesn't proceed to run the program.

Example script:

```
#Listener
async call comms

:comms
listen "ReferenceCoords"
wait until false
repeat

#Sender

:TransmitReference
set Command to "set global reference to " + ("Remote Control" position) + " as vector"
print Command
send Command to "ReferenceCoords"
repeat
```

I believe the fix here is to wipe all channels that were previously being listened to whenever we reset the program state.

This means the script will ignore any previous messages sent when you re-compile, which is appropriate.  It would fix other bugs too such as changing the channel you are listening to, which currently wouldn't cause you to stop listening on the old channel I believe.

MerlinofMines:
…g script

Previously broadcast listeners would incorrectly remain active even after re-parsing the script.

This commit also removes an unnecessary delegate from the script whose purpose was only to give us a hook for setting messages.  This has been replaced with a
complete Mock IGC instance.

This PR fixes #167

### DIFF
--- a/EasyCommands/Program.cs
+++ b/EasyCommands/Program.cs
@@ -33,10 +33,7 @@ namespace IngameScript {
         public int maxItemTransfers = 10;
         #endregion
 
-        public delegate List<MyIGCMessage> BroadcastMessageProvider();
-
         public static Program PROGRAM;
-        public BroadcastMessageProvider broadcastMessageProvider;
         public ProgramState state = ProgramState.STOPPED;
         public Dictionary<String, FunctionDefinition> functions = NewDictionary<string, FunctionDefinition>();
         public Thread currentThread;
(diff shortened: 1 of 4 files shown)


EasyCommands/Program.cs
```diff
@@ -58,6 +55,15 @@ public partial class Program : MyGridProgram {
         List<String> commandStrings = NewList<String>();
 
         public void ClearAllState() {
+            var listeners = NewList<IMyBroadcastListener>();
+            IGC.GetBroadcastListeners(listeners);
```
MerlinofMines, line 59: On second look, a function providing the broadcast listeners may be less chars

PR #188, opened 2022-01-30
Dynamic item types

### LINKED ISSUE #148: Dynamic ItemName Parsing
The current Item Name parsing for inventories and production queues relies on hardcoded mappings between specific names and in game Blueprint/Item names.

As such, only vanilla items are currently supported for inventory management. We should revisit this and see if there is a way to dynamically resolve these such that 3P can also be used.

As part of this, we might also be able to add support for printing out the needed component items for any block, as well as print out the total components needed for finishing a blueprint from a projector, which would be handy

MerlinofMines:
This PR adds 2 commits for Dynamic Item support.  First, it adds the ability to parse and use custom TypeIds and SubTypesIds, both for inventories and for blueprints.

Second, it adds a new Types property for inventories and assemblers for printing out TypeIds, SubTypeIds, and BlueprintIds for custom items.

This PR resolves #148


EasyCommands/CommandParsers/ParameterParsers.cs
```diff
@@ -101,6 +101,7 @@ partial class Program {
             AddPropertyWords(Words("use", "used", "occupy", "occupied", "control", "controlled"), Property.USE);
             AddPropertyWords(Words("unused", "unoccupied", "vacant", "available"), Property.USE, false);
             AddPropertyWords(Words("done", "ready", "complete", "finished", "built", "finish", "pressurized", "depressurized"), Property.COMPLETE);
+            AddPropertyWords(Words("clear", "wipe", "erase"), Property.COMPLETE, false);
```
MerlinofMines, line 104: Need to update docs for these words

PR #191, opened 2022-01-31
Added support for Dynamic Function Execution

### LINKED ISSUE #172: Dynamic Function Calls
It'd be nice if you could dynamically invoke a function using ```call variable``` or ```goto variable```

This would require some new parsing rules.  Also, safety on the function name as well as parameter count would have to be best effort.

Note that we may have to relax some of the current function definition validation for this, as the function being invoked won't be known until runtime.

MerlinofMines:
This commit adds the ability to invoke functions dynamically using variables.  This is a powerful feature, but comes at the cost of compile time safety.

This commit moves the resolution of the function name to invoke to a runtime call.  It also leaves the resolution of parameter key names to runtime.  The existing
error messages have been improved, but will now show up as runtime exceptions.

Lastly, this commit introduces a "Utilities" section to contain some useful utilities for saving characters on Type and Function declarations.

This commit resolves #172

### DIFF
--- a/docs/EasyCommands/functions.md
+++ b/docs/EasyCommands/functions.md
@@ -59,6 +59,37 @@ goto "myFunction2"
 
 This is effectively changing what function is currently considered "active" by the program.  The current function, no matter its state of execution, is abandoned in favor of the new active program.  Furthermore, any commands to "restart", "loop", "start" will now restart the new function.  This behavior effectively enables "modes" for your program to run in where the modes are intended to be exclusive. 
 
+### Calling Functions Dynamically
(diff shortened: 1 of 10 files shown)


docs/EasyCommands/functions.md
```diff
@@ -59,6 +59,37 @@ goto "myFunction2"
 
 This is effectively changing what function is currently considered "active" by the program.  The current function, no matter its state of execution, is abandoned in favor of the new active program.  Furthermore, any commands to "restart", "loop", "start" will now restart the new function.  This behavior effectively enables "modes" for your program to run in where the modes are intended to be exclusive. 
 
+### Calling Functions Dynamically
+It is possible to invoke a named function using a variable specifying the name of the function to invoke by explicitly using ```call``` or ```goto```.  This provides an easy way to create modes for your program.  Make sure to pass the same number of parameters to the function you intend to invoke.
+
+```
+set global runningFunction to to "attack"
```
MerlinofMines, line 66: nit: to to -> to

PR #193, opened 2022-02-01
Small Changes

jgersti:
* add `round` operator for vectors
* move some function from Collections.cs to Utilities.cs
* use gyro properties directly again and introduce constant for conversion from RPM to rad/s and viceversa
    each vector component is seperatly multiplied by the conversion factors so the test does not fail because of float to double promotion.

### DIFF
--- a/EasyCommands/BlockHandlers/GyroscopeBlockHandlers.cs
+++ b/EasyCommands/BlockHandlers/GyroscopeBlockHandlers.cs
@@ -29,15 +29,15 @@ namespace IngameScript {
 
                 var overrideHandler = DirectionalTypedHandler(Direction.NONE,
                         TypeHandler(ReturnTypedHandler(Return.VECTOR,
-                            TypeHandler(VectorHandler(b => new Vector3D(b.Pitch * RadiansPerSecToRPM, b.Yaw * RadiansPerSecToRPM, b.Roll * RadiansPerSecToRPM), (b, v) => {
-                                b.Pitch = RPMToRadiansPerSec * (float)v.X;
+                            TypeHandler(VectorHandler(b => new Vector3D(-b.Pitch * RadiansPerSecToRPM, b.Yaw * RadiansPerSecToRPM, b.Roll * RadiansPerSecToRPM), (b, v) => {
+                                b.Pitch = RPMToRadiansPerSec * (float)-v.X;
                                 b.Yaw = RPMToRadiansPerSec * (float)v.Y;
                                 b.Roll = RPMToRadiansPerSec * (float)v.Z;
                             }), Return.VECTOR),
                             TypeHandler(BooleanHandler(b => b.GyroOverride, (b, v) => b.GyroOverride = v), Return.BOOLEAN))
                         , Direction.NONE),
-                        TypeHandler(NumericHandler(b => RadiansPerSecToRPM * b.Pitch, (b, v) => b.Pitch = RPMToRadiansPerSec * v, 5), Direction.UP),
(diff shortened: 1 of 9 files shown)


EasyCommands/BlockHandlers/GyroscopeBlockHandlers.cs
```diff
@@ -29,19 +29,19 @@ public class GyroscopeBlockHandler<T> : FunctionalBlockHandler<T> where T : clas
 
                 var overrideHandler = DirectionalTypedHandler(Direction.NONE,
                         TypeHandler(ReturnTypedHandler(Return.VECTOR,
-                            TypeHandler(VectorHandler(b => new Vector3D(GetPitch(b), GetYaw(b), GetRoll(b)), (b, v) => {
-                                SetPitch(b, (float)v.X);
-                                SetYaw(b, (float)v.Y);
-                                SetRoll(b, (float)v.Z);
+                            TypeHandler(VectorHandler(b => new Vector3D(b.Pitch * RadiansPerSecToRPM, b.Yaw * RadiansPerSecToRPM, b.Roll * RadiansPerSecToRPM), (b, v) => {
```
MerlinofMines, line 32: nit: I'm pretty sure you can pull out the multiplication:
```RadiansPerSecToRPM * new Vector3D(b.Pitch, b.Yaw, b.Roll)```

**Comment by MerlinofMines** [comment]
So, testing locally I don't like what the conversion does to the values.

An ingame value of "1" gets converted to "0.9999995".  I think this will end up being annoying to most folks trying to use the properties.

I think it may be better to leave as is.

File: EasyCommands/BlockHandlers/GyroscopeBlockHandlers.cs (hunk 1 above)
Comment (MerlinofMines) [inline EasyCommands/BlockHandlers/GyroscopeBlockHandlers.cs:32]: Testing in game, it looks like Pitch is inverted from what you'd expect.  1.0 Overrides = -1.0 when you call b.Pitch.  I think we'll need to invert this.

Comment (jgersti) [reply EasyCommands/BlockHandlers/GyroscopeBlockHandlers.cs:32]: that is indeed possible but does lead to problems with the test since the numbers are not truncated to float again and so do not match when the multiplication is outside.

Comment (jgersti) [reply EasyCommands/BlockHandlers/GyroscopeBlockHandlers.cs:32]: Looking at the code pitch should be correct, the other two might not actually do the (mathematically) expected things.

Following the right hand rule these should be
* `x > 0` pitch up
* `y > 0` yaw left
*  `z > 0` roll ccw
but i think the game inverts yaw and roll so they are right and cw

Comment (jgersti) [reply EasyCommands/BlockHandlers/GyroscopeBlockHandlers.cs:32]: Since i cannot find the comment with accuraccy again i will reply here . with the current order of operation the output should be exactly the same as before.

But this is a general problem with vectors being double and numbers being float internally.

Comment (jgersti) [reply EasyCommands/BlockHandlers/GyroscopeBlockHandlers.cs]: The coordinate systems in this game are seriously messsed up...

You where right pitch gets inverted and now i understand why you inverted the mouse input aswell.

With a correctly oriented gyro this script now mimcs the normal controls (just more sluggish)
```
async TestRotationControls

:TestRotationControls
set my gyros to true
set my gyros to my cockpit rotation
replay
```

MerlinofMines: Indeed, I've considered changing the internals to use doubles.  I only did float because most properties in SE were float so it was less chars, lol.

MerlinofMines (approved):

PR #195, opened 2022-02-01
Attach/Detach support for suspensions and rotors

### LINKED ISSUE #183: Attach/Detach Support
Currently Pistons and Wheels lack attach/detach support, even though there is an action for this on these blocks.  Perhaps we can enable this using the Connected property?

As an action, the attach/detach action would have to be best effort

```
while "My Wheel" is not attached
  attach "My Wheel"
```

jgersti:
closes #183.

The Clang disclaimer might need to be worded more strongly.


MerlinofMines:
I'm happy with it!

PR #218, opened 2022-02-13
Cache Block and Group retrieval

jgersti:
`BlockHandlerRegistry` needs some changes i guess and i reshuffled `BlockHandlers.cs` a bit.

### DIFF
--- a/EasyCommands/BlockHandlers/BlockHandlers.cs
+++ b/EasyCommands/BlockHandlers/BlockHandlers.cs
@@ -10,6 +10,11 @@ using VRageMath;
 
 namespace IngameScript {
     partial class Program {
+        public List<IMyTerminalBlock> blockCache = NewList<IMyTerminalBlock>();
+        public List<IMyBlockGroup> groupCache = NewList<IMyBlockGroup>();
+        public bool invalidBlockCache = true;
+        public bool invalidGroupCache = true;
+
         public static class BlockHandlerRegistry {
             static Dictionary<Block, BlockHandler> blockHandlers = new Dictionary<Block, BlockHandler> {
                 { Block.AIRVENT, new AirVentBlockHandler()},
@@ -71,9 +76,23 @@ namespace IngameScript {
                 return blockHandlers[blockType];
             }
 
-            public static List<Object> GetBlocks(Block blockType, Func<IMyTerminalBlock, bool> selector) => blockHandlers[blockType].GetBlocks(selector);
+            public static List<Object> GetBlocks(Block blockType, Func<IMyTerminalBlock, bool> selector = null) {
+                if (PROGRAM.invalidBlockCache) {
+                    PROGRAM.GridTerminalSystem.GetBlocks(PROGRAM.blockCache);
+                    PROGRAM.invalidBlockCache = false;
+                }
+                return blockHandlers[blockType].SelectBlocks(PROGRAM.blockCache, selector);
+            }
 
-            public static List<Object> GetBlocksInGroup(Block blockType, String groupName) => blockHandlers[blockType].GetBlocksInGroup(groupName);
+            public static List<Object> GetBlocksInGroup(Block blockType, String groupName) {
+                if (PROGRAM.invalidGroupCache) {
+                    PROGRAM.GridTerminalSystem.GetBlockGroups(PROGRAM.groupCache);
+                    PROGRAM.invalidGroupCache = false;
+                }
+                var blocks = NewList<IMyTerminalBlock>();
+                PROGRAM.groupCache.Find(g => g.Name == groupName)?.GetBlocks(blocks);
+                return blockHandlers[blockType].SelectBlocks(blocks);
+            }
         }
 
         //Property Getters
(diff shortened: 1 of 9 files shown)


EasyCommands/BlockHandlers/BlockHandlers.cs
```diff
@@ -71,9 +76,34 @@ public static class BlockHandlerRegistry {
...
+                return PROGRAM.blockCache;
+            }
+
+            public static List<IMyBlockGroup> RetrieveGroups() {
+                if (PROGRAM.invalidGroupCache) {
+                    PROGRAM.GridTerminalSystem.GetBlockGroups(PROGRAM.groupCache);
+                    PROGRAM.invalidGroupCache = false;
+                }
+                return PROGRAM.groupCache;
+            }
```
MerlinofMines, line 99: nit: I believe we can inline these into GetBLocks and GetBlocksInGroup, respectively

EasyCommands/BlockHandlers/BlockHandlers.cs
```diff
@@ -71,9 +76,34 @@ public static class BlockHandlerRegistry {
                 return blockHandlers[blockType];
             }
 
-            public static List<Object> GetBlocks(Block blockType, Func<IMyTerminalBlock, bool> selector) => blockHandlers[blockType].GetBlocks(selector);
+            public static void invalidateCaches() {
+                PROGRAM.invalidBlockCache = true;
+                PROGRAM.invalidGroupCache = true;
+            }
```
MerlinofMines, line 82: nit let's inline these inside of Program.  Can save "public static void m() {}" (25 chars)

EasyCommands/BlockHandlers/BlockHandlers.cs
```diff
@@ -10,6 +10,11 @@
 
 namespace IngameScript {
     partial class Program {
+        public List<IMyTerminalBlock> blockCache = NewList<IMyTerminalBlock>();
+        public List<IMyBlockGroup> groupCache = NewList<IMyBlockGroup>();
+        public bool invalidBlockCache = true;
+        public bool invalidGroupCache = true;
```
MerlinofMines, line 16: you *might* be able to get away with a null check against blockCache/groupCache instead?

EasyCommands/BlockHandlers/BlockHandlers.cs
```diff
@@ -315,9 +324,11 @@ public abstract class BlockHandler<T> : BlockHandler where T : class {
             PropertyHandler<T> TypedPropertyHandler<U>(GetTypedProperty<T, U> Get, SetTypedProperty<T, U> Set, Func<Primitive, U> Cast, U delta) => new SimpleTypedHandler<T, U>(Get, Set ?? ((b, v) => { }), Cast, delta);
         }
 
-        public class TypeHandler<T, U> {
-            public PropertyHandler<T> handler;
-            public U[] supportedTypes;
+        public abstract class MultiInstanceBlockHandler<T> : BlockHandler<T> where T : class {
+            public override IEnumerable<T> SelectBlocksByType<U>(List<U> blocks, Func<U, bool> selector = null) =>
+                blocks?.Where(b => selector?.Invoke(b) ?? true).SelectMany(b => GetInstances(b)) ?? Enumerable.Empty<T>();
```
MerlinofMines, line 329: I wonder if we should add a List<T>(IEnumerable<T> things) utility.  At some point it may be less chars. "IEnumerable" vs "List", especially given NewList<T>(), causes chars to add up for sure.  7 chars diff per Enumerable vs List.

EasyCommands/BlockHandlers/BlockHandlers.cs
jgersti: no

EasyCommands/BlockHandlers/BlockHandlers.cs
jgersti: you need to count the more or less mandatory `.ToList()` when using lists aswell which ats 9 more chars.
The advantage of IEnumerable is that the traversal is deferred until it is realized (e.g. evtl calling to list)

EasyCommands/BlockHandlers/BlockHandlers.cs
jgersti: sure. i just not sure where to put theis stuff. The beginning of main seems inapppropriate.

MerlinofMines: i think the beginning of "RunThreads" might be reasonable

jgersti: sounds better than the beginning of main

EasyCommands/BlockHandlers/BlockHandlers.cs
```diff
@@ -71,9 +76,29 @@ public static class BlockHandlerRegistry {
...
+            }
+
+            public static List<Object> GetBlocks(Block blockType, Func<IMyTerminalBlock, bool> selector = null) {
+                if (PROGRAM.invalidBlockCache) {
+                    PROGRAM.GridTerminalSystem.GetBlocks(PROGRAM.blockCache);
+                    PROGRAM.invalidBlockCache = false;
+                }
+                return blockHandlers[blockType].SelectBlocks(PROGRAM.blockCache, selector);
+            }
+            public static List<Object> GetBlocks(Block blockType, string selectorString) => GetBlocks(blockType, b => b.CustomName.Equals(selectorString));
```
MerlinofMines, line 91: nit: I think we only use this for one call.  I suggest we inline and just use the lambda:
```b => b.CustomName.Equals(selectorString)```

Saves a decent chunk of chars

File: EasyCommands/BlockHandlers/TerminalBlockHandlers.cs [hunk 6]
```diff
@@ -94,21 +86,9 @@ public class TerminalBlockHandler<T> : BlockHandler<T> where T : class, IMyTermi
...
-                var blockGroups = NewList<IMyBlockGroup>();
-                PROGRAM.GridTerminalSystem.GetBlockGroups(blockGroups);
-                IMyBlockGroup group = blockGroups.Find(g => g.Name.Equals(groupName));
-                var blocks = NewList<T>();
-                if (group == null) return blocks;
-                group.GetBlocksOfType<T>(blocks);
-                return blocks;
-            }
+            public override IEnumerable<T> SelectBlocksByType<U>(List<U> blocks, Func<U, bool> selector = null) =>
+                blocks?.Where(b => selector?.Invoke(b) ?? true).OfType<T>() ?? Enumerable.Empty<T>();
```
MerlinofMines, line 91: Do we need the null check against blocks and the fallback ENumerable.Empty<T>()?  I don't think we ever call this method with null blocks, do we?

EasyCommands/BlockHandlers/TerminalBlockHandlers.cs
```diff
@@ -144,6 +124,13 @@ public class TerminalBlockHandler<T> : BlockHandler<T> where T : class, IMyTermi
 
             public PropertyHandler<T> TerminalBlockPropertyHandler(String propertyId, object delta) => new TerminalBlockPropertyHandler<T>(propertyId, ResolvePrimitive(delta));
         }
+        public class FunctionalBlockHandler<T> : TerminalBlockHandler<T> where T : class, IMyFunctionalBlock {
+            public FunctionalBlockHandler() : base() {
```
MerlinofMines, line 128: nit: I think we can remove :base()

EasyCommands/BlockHandlers/BlockHandlers.cs
jgersti: that function should have served as the hook in point for memoization. But since i have no idea how to do that efficitantly right now it is not needed

EasyCommands/BlockHandlers/TerminalBlockHandlers.cs
jgersti: maybe i was a bit paranoid there ...

PR #219, opened 2022-02-15
Memoization 

jgersti:
- [x] cleanup `Selector` implementations
- [x] cleanup `BlockHandler` implementations
- [x] implement storage for caches (`Cache`)
- [x] cache block retrieval for block for block type + selector string per tick
- [x] cache block retrieval for block for block type + group name per tick
- [x] cache `ITerminalAction`s
- [x] cache `ITerminalProperties`s 
- ~~[ ] move `BlockhandlerRegistry` and caches into a seperate file?~~

### DIFF
--- a/EasyCommands/BlockHandlers/BlockHandlers.cs
+++ b/EasyCommands/BlockHandlers/BlockHandlers.cs
@@ -12,8 +12,6 @@ namespace IngameScript {
     partial class Program {
         public List<IMyTerminalBlock> blockCache = NewList<IMyTerminalBlock>();
         public List<IMyBlockGroup> groupCache = NewList<IMyBlockGroup>();
-        public bool invalidBlockCache = true;
-        public bool invalidGroupCache = true;
 
         public static class BlockHandlerRegistry {
             static Dictionary<Block, BlockHandler> blockHandlers = new Dictionary<Block, BlockHandler> {
@@ -76,19 +74,21 @@ namespace IngameScript {
                 return blockHandlers[blockType];
             }
 
-            public static List<Object> GetBlocks(Block blockType, Func<IMyTerminalBlock, bool> selector = null) {
-                if (PROGRAM.invalidBlockCache) {
+            public static List<Object> GetSelf(Block? blockType) =>
+                blockType == null || blockType == Block.DISPLAY
+                    ? blockHandlers[blockType ?? Block.PROGRAM].SelectBlocks(NewList(PROGRAM.Me))
+                    : null;
+
+            public static List<Object> GetBlocks(Block blockType, string selector = null) {
+                if (PROGRAM.blockCache.Count == 0)
                     PROGRAM.GridTerminalSystem.GetBlocks(PROGRAM.blockCache);
-                    PROGRAM.invalidBlockCache = false;
-                }
-                return blockHandlers[blockType].SelectBlocks(PROGRAM.blockCache, selector);
+                return blockHandlers[blockType].SelectBlocks(PROGRAM.blockCache, b => selector?.Equals(b.CustomName) ?? true);
             }
 
             public static List<Object> GetBlocksInGroup(Block blockType, String groupName) {
-                if (PROGRAM.invalidGroupCache) {
+                if (PROGRAM.groupCache.Count == 0)
                     PROGRAM.GridTerminalSystem.GetBlockGroups(PROGRAM.groupCache);
-                    PROGRAM.invalidGroupCache = false;
-                }
+
                 var blocks = NewList<IMyTerminalBlock>();
                 PROGRAM.groupCache.Find(g => g.Name == groupName)?.GetBlocks(blocks);
                 return blockHandlers[blockType].SelectBlocks(blocks);
@@ -139,7 +139,7 @@ namespace IngameScript {
                 Reverse = (b, p) => Set(b, p, Get(b, p).Not());
             }
 
-            Primitive Multiply(Primitive p, PropertySupplier property) => property.increment.GetValueOrDefault(true) ? p : p.Not();
+            Primitive Multiply(Primitive p, PropertySupplier property) => property.increment ?? true ? p : p.Not();
         }
 
         public class SimpleTypedHandler<T, U> : SimplePropertyHandler<T> {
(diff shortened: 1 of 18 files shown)


jgersti:
I did a quick test with the skid stering script the avg runtime and instruction count is ~25% down.

EasyCommands/Commands/Selectors.cs
```diff
@@ -49,27 +49,21 @@ public class IndexSelector : Selector {
...
+                    .Select(v => v.GetValue())
+                    .SelectMany(p => {
+                        if (p.returnType == Return.NUMERIC)
+                            return entities.GetRange((int)CastNumber(p), 1);
+                        else if (p.returnType == Return.STRING) {
+                            var s = CastString(p);
+                            return entities.Where(o => s == b.GetName(o));
+                        } else
+                            return NewList<Object>();
+                    }).ToList();
```
MerlinofMines, line 66: may be worth creating a variable outside this and remove the ```else return NewList<Object>()```?

EasyCommands/BlockHandlers/BlockHandlers.cs
```diff
@@ -76,23 +76,25 @@ public static class BlockHandlerRegistry {
                 return blockHandlers[blockType];
             }
 
-            public static List<Object> GetBlocks(Block blockType, Func<IMyTerminalBlock, bool> selector = null) {
-                if (PROGRAM.invalidBlockCache) {
+            public static List<Object> GetSelf(Block? blockType) =>
+                blockType == null || blockType == Block.DISPLAY
+                    ? blockHandlers[blockType ?? Block.PROGRAM].SelectBlocks(NewList(PROGRAM.Me))
+                    : null;
```
MerlinofMines, line 82: This is breaking the behavior of SelfSelector, which should return all blocks of the given type (no selector name) if the block type is anything but Display, Program or null.

jgersti: the `SelfSelector` calls `BGR.GetSelf(blockType) ?? BGR.GetBlocks(GetBlockType())`

EasyCommands/BlockHandlers/BlockHandlers.cs
```diff
@@ -76,23 +76,25 @@ public static class BlockHandlerRegistry {
...
-                PROGRAM.groupCache.Find(g => g.Name == groupName)?.GetBlocks(blocks);
-                return blockHandlers[blockType].SelectBlocks(blocks);
+                return PROGRAM.selectorCache.GetOrCreate(MyTuple.Create(blockType, selector), () =>
+                    blockHandlers[blockType].SelectBlocks(PROGRAM.blockCache, b => selector?.Equals(b.CustomName) ?? true));
             }
+
+            public static List<Object> GetBlocksInGroup(Block blockType, String groupName) =>
+                PROGRAM.groupCache.GetOrCreate(MyTuple.Create(blockType, groupName), () => {
+                    var blocks = NewList<IMyTerminalBlock>();
+                    PROGRAM.GridTerminalSystem.GetBlockGroupWithName(groupName)?.GetBlocks(blocks);
```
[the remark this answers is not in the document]
jgersti: but yes that is approximatly what i planned.

EasyCommands/Program.cs
```diff
@@ -160,8 +160,9 @@ public partial class Program : MyGridProgram {
         void RunThreads() {
             try {
                 // invalidate block & group cache
-                invalidBlockCache = true;
-                invalidGroupCache = true;
+                blockCache.Clear();
+                groupCache.Clear();
+                selectorCache.Clear();
```
MerlinofMines, line 165: Clever that we aren't clearing the actions/properties.  This would definitely improve script performance when using actions/properties directly.

EasyCommands/BlockHandlers/TerminalBlockHandlers.cs
```diff
@@ -19,38 +19,41 @@
...
+                KeyValuePair("Color", PropertyConverter((p, b) => p.AsColor().GetValue(b), (p, b, v) => p.AsColor().SetValue(b, CastColor(v))))
             );
 
-            static TerminalPropertyConverter GetPropertyConverter(T block, PropertySupplier propertySupplier) {
-                var property = block.GetProperty(propertySupplier.propertyType);
-                if (property == null) throw new Exception(block.BlockDefinition.SubtypeName + " does not have property: " + (propertySupplier.propertyWord ?? propertySupplier.propertyType));
-                return TerminalPropertyValueConversion[property.TypeName];
-            }
+            static ITerminalProperty GetTerminalProperty(T block, PropertySupplier propertySupplier) =>
+                PROGRAM.propertyCache.GetOrCreate(typeof(T), propertySupplier.propertyType, s =>{
```
MerlinofMines, line 51: Are we sure that the properties for a given block are always the same when queried by class type?  For example, the Projector and Console Block have the same block type.  Are we sure the properties will be the same?

I'm a little concerned about use cases where we would query a sub block as a "IMyTerminalBlock" block, and then cache a property by type for IMyTerminalBlock incorrectly.  

Example:

```
set "My Beacon" terminal "Range" property to 200
set "My Gatling Turret" terminal "Range" property to 200
```

In both cases we are caching the "Range" property for the IMyTerminalBlock (I believe) but the property for the individual blocks is actually different.

jgersti: Terminal properties and actions are only defined by block type. So that is not the problem.
But as you pointed out he code above might do the wrong thing. Replacing `typeof(T)` by `b.GetType()` will get the `Type` instance for the actual block type.

EasyCommands/Program.cs
jgersti: we might want think about keeping (some of) these caches alive for multiple ticks. Every 10th tick would still refresh the caches six times per second.

MerlinofMines: I think we hold off on that one for now.  Users can change the run frequency and there's no guarantee that the program is running constantly (buttons to do commands), so the caches would have to be refreshed based on a "last updated time".

I think caching per tick is fine for now.

EasyCommands/BlockHandlers/TerminalBlockHandlers.cs
MerlinofMines: Ya the b.GetType() should be more accurate.

Let's add a few tests to make sure that caching is working as so we don't break it in the future somehow.

MerlinofMines (approved):

MerlinofMines (approved):

PR #223, opened 2022-02-22
Refactoring

jgersti:
This is a edit (and corrected) version of some general purpose refactoring and char saving measures.
This accomplished by heavy usage of the operators `?.` and `??` and removal of obsolete or unnecessary code.

edit:
master: 87374 chars
Refactor: 85584 chars

### DIFF
--- a/EasyCommands/BlockHandlers/TextSurfaceHandlers.cs
+++ b/EasyCommands/BlockHandlers/TextSurfaceHandlers.cs
@@ -64,10 +64,10 @@ namespace IngameScript {
 
             public override IEnumerable<IMyTextSurface> GetInstances(IMyTerminalBlock block) =>
                 block is IMyTextSurface
-                    ? Enumerable.Repeat((IMyTextSurface)block, 1)
+                    ? Once((IMyTextSurface)block)
                     : block is IMyTextSurfaceProvider
-                        ? Enumerable.Range(0, ((IMyTextSurfaceProvider)block).SurfaceCount).Select(((IMyTextSurfaceProvider)block).GetSurface)
-                        : Enumerable.Empty<IMyTextSurface>();
+                        ? Range(0, ((IMyTextSurfaceProvider)block).SurfaceCount).Select(((IMyTextSurfaceProvider)block).GetSurface)
+                        : Empty<IMyTextSurface>();
         }
     }
 }
--- a/EasyCommands/BlockHandlers/TextSurfaceHandlers.cs
+++ b/EasyCommands/BlockHandlers/TextSurfaceHandlers.cs
@@ -34,7 +34,7 @@ namespace IngameScript {
                 AddListHandler(Property.MEDIA_LIST, b => {
                     var images = NewList<string>();
                     b.GetSelectedImages(images);
-                    return NewKeyedList(images.Select(i => GetStaticVariable(i)));
+                    return NewKeyedList(images.Select(GetStaticVariable));
                 }, SetImages);
                 AddStringHandler(Property.POSITION, b => (b.Alignment + "").ToLower(), (b, v) => b.Alignment = v == "center" ? TextAlignment.CENTER : v == "right" ? TextAlignment.RIGHT : TextAlignment.LEFT);
                 AddNumericHandler(Property.INTERVAL, b => b.ChangeInterval, (b, v) => b.ChangeInterval = v, 1);
@@ -63,11 +63,11 @@ namespace IngameScript {
             public override string Name(IMyTextSurface block) => block.DisplayName;
 
             public override IEnumerable<IMyTextSurface> GetInstances(IMyTerminalBlock block) =>
-                block is IMyTextSurface ?
-                Enumerable.Repeat((IMyTextSurface)block, 1) :
-                block is IMyTextSurfaceProvider ?
-                Enumerable.Range(0, ((IMyTextSurfaceProvider)block).SurfaceCount).Select(i => ((IMyTextSurfaceProvider)block).GetSurface(i)) :
-                Enumerable.Empty<IMyTextSurface>();
+                block is IMyTextSurface
+                    ? Enumerable.Repeat((IMyTextSurface)block, 1)
+                    : block is IMyTextSurfaceProvider
+                        ? Enumerable.Range(0, ((IMyTextSurfaceProvider)block).SurfaceCount).Select(((IMyTextSurfaceProvider)block).GetSurface)
+                        : Enumerable.Empty<IMyTextSurface>();
         }
     }
 }
--- a/EasyCommands/CommandParsers/ParameterParsers.cs
+++ b/EasyCommands/CommandParsers/ParameterParsers.cs
@@ -20,7 +20,7 @@ using VRageMath;
 namespace IngameScript {
     partial class Program {
         //Internal (Don't touch!)
-        Dictionary<String, List<CommandParameter>> propertyWords = NewDictionary<string, List<CommandParameter>>();
+        Dictionary<String, List<ICommandParameter>> propertyWords = NewDictionary<string, List<ICommandParameter>>();
 
         string[] firstPassTokens = new[] { "(", ")", "[", "]", ",", "*", "/", "!", "^", "..", "%", ">=", "<=", "==", "&&", "||", "@", "$", "->", "++", "+=", "--", "-=", "::"};
         string[] secondPassTokens = new[] { "<", ">", "=", "&", "|", "-", "+", "?", ":" };
@@ -402,25 +402,25 @@ namespace IngameScript {
             AddWords(groupWords, new BlockTypeCommandParameter(blockType), new GroupCommandParameter());
         }
 
-        void AddAmbiguousWords(String[] words, params CommandParameter[] commandParameters) {
+        void AddAmbiguousWords(String[] words, params ICommandParameter[] commandParameters) {
             foreach (String word in words)
                 AddWords(Words(word), new AmbiguousStringCommandParameter(word, false, new AmbiguousCommandParameter(commandParameters)));
         }
 
-        void AddWords(String[] words, params CommandParameter[] commandParameters) {
+        void AddWords(String[] words, params ICommandParameter[] commandParameters) {
             foreach (String word in words) propertyWords.Add(word, commandParameters.ToList());
         }
 
-        List<CommandParameter> ParseCommandParameters(List<Token> tokens) => tokens.SelectMany(t => ParseCommandParameters(t)).ToList();
+        List<ICommandParameter> ParseCommandParameters(List<Token> tokens) => tokens.SelectMany(t => ParseCommandParameters(t)).ToList();
 
-        List<CommandParameter> ParseCommandParameters(Token token) {
-            var commandParameters = NewList<CommandParameter>();
+        List<ICommandParameter> ParseCommandParameters(Token token) {
+            var commandParameters = NewList<ICommandParameter>();
             String t = token.token;
             if (token.isExplicitString) {
                 commandParameters.Add(new VariableCommandParameter(GetStaticVariable(token.original)));
             } else if (token.isString) {
                 List<Token> subTokens = Tokenize(t);
-                List<CommandParameter> subtokenParams = ParseCommandParameters(subTokens);
+                List<ICommandParameter> subtokenParams = ParseCommandParameters(subTokens);
                 commandParameters.Add(new AmbiguousStringCommandParameter(token.original, false, subtokenParams.ToArray()));
             } else if (propertyWords.ContainsKey(t)) {
                 commandParameters.AddList(propertyWords[t]);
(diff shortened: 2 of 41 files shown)


EasyCommands/BlockHandlers/TextSurfaceHandlers.cs
```diff
@@ -63,11 +63,11 @@ public class TextSurfaceHandler : MultiInstanceBlockHandler<IMyTextSurface> {
             public override string Name(IMyTextSurface block) => block.DisplayName;
 
             public override IEnumerable<IMyTextSurface> GetInstances(IMyTerminalBlock block) =>
-                block is IMyTextSurface ?
-                Enumerable.Repeat((IMyTextSurface)block, 1) :
-                block is IMyTextSurfaceProvider ?
-                Enumerable.Range(0, ((IMyTextSurfaceProvider)block).SurfaceCount).Select(i => ((IMyTextSurfaceProvider)block).GetSurface(i)) :
-                Enumerable.Empty<IMyTextSurface>();
+                block is IMyTextSurface
+                    ? Once((IMyTextSurface)block)
```
MerlinofMines, line 67: nit: I think you could use NewList() here to save a few chars

EasyCommands/BlockHandlers/TextSurfaceHandlers.cs
```diff
@@ -63,11 +63,11 @@ public class TextSurfaceHandler : MultiInstanceBlockHandler<IMyTextSurface> {
...
+                    : block is IMyTextSurfaceProvider
+                        ? Range(0, ((IMyTextSurfaceProvider)block).SurfaceCount).Select(((IMyTextSurfaceProvider)block).GetSurface)
+                        : Empty<IMyTextSurface>();
```
MerlinofMines, line 70: nit: NewList<IMyTextSurface>()

EasyCommands/CommandParsers/ParameterParsers.cs
```diff
@@ -444,30 +444,28 @@ partial class Program {
...
-        Token[] TokenizeEnclosed(String token, string characters, Func<String, Token[]> parseSubTokens) =>
+        IEnumerable<Token> TokenizeEnclosed(String token, string characters, Func<String, IEnumerable<Token>> parseSubTokens) =>
             characters.Length == 0 ? parseSubTokens(token) :
                 token.Trim().Split(characters[0])
                 .SelectMany((element, index) => index % 2 == 0  // If even index
-                    ? TokenizeEnclosed(element, characters.Remove(0,1), parseSubTokens)  // Split the item
-                    : new Token[] { new Token(element, true, characters.Length > 1) })  // Keep the entire item
-                .ToArray();
+                    ? TokenizeEnclosed(element, characters.Remove(0, 1), parseSubTokens)  // Split the item
+                    : Once(new Token(element, true, characters.Length > 1)));  // Keep the entire item
```
MerlinofMines, line 456: I see Once() several times. Would NewList() work here to save 3 chars per usage?

EasyCommands/BlockHandlers/TextSurfaceHandlers.cs
MerlinofMines: Rather, why is the Once method necessary, given ```NewList(item)``` ?

EasyCommands/Common/Utilities.cs
```diff
@@ -30,10 +30,16 @@ partial class Program {
...
+        public static IEnumerable<int> Range(int start, int end) => Enumerable.Range(start, end);
+        public static IEnumerable<T> Empty<T>() => Enumerable.Empty<T>();
+        public static IEnumerable<T> Once<T>(T element) => Enumerable.Repeat(element, 1);
 
         //Other useful utilities
-        public static Variable GetStaticVariable(object o) => new StaticVariable(ResolvePrimitive(o));
+        public static IVariable GetStaticVariable(object o) => new StaticVariable(ResolvePrimitive(o));
+        public static IVariable EmptyList() => GetStaticVariable(NewKeyedList());
         public static Vector3D Vector(double x, double y, double z) => new Vector3D(x, y, z);
+        public static bool AnyNotNull(params Object[] objects) => objects.Any(o => o != null);
```
MerlinofMines, line 43: We only use this method a couple of times, is it less chars than checking the nulls directly?

EasyCommands/Commands/Commands.cs
```diff
@@ -260,27 +260,27 @@ public class SendCommand : Command {
...
                 return true;
             }
         }
 
         public class TransferItemCommand : Command {
-            public Selector from;//Must be Inventory
-            public Selector to;//Must be Inventory
-            public Variable first, second;//One of these is an amount (nullable), other must be ItemFilter (non nullable)
+            public ISelector from;//Must be Inventory
+            public ISelector to;//Must be Inventory
```
MerlinofMines, line 280: nit: ```public ISelector from, to;```

EasyCommands/Commands/Commands.cs
```diff
@@ -412,7 +410,7 @@ public class MultiActionCommand : Command {
                     }
                 }
 
-                if (currentCommands != null && currentCommands.Count > 0) return false;
+                if (currentCommands?.Count > 0) return false;
```
MerlinofMines, line 413: is the ``?? 0``` needed here as well? Otherwise, is it not needed above?

EasyCommands/Common/Aggregators.cs
```diff
@@ -20,6 +20,6 @@
 namespace IngameScript {
     partial class Program {
         public delegate Primitive Aggregator(IEnumerable<object> blocks, Func<object,Primitive> primitiveSupplier);
-        public Aggregator SumAggregator = (blocks, primitiveSupplier) => blocks.Select(b => primitiveSupplier(b)).DefaultIfEmpty(ResolvePrimitive(0)).Aggregate((a, b) => a.Plus(b));
+        public Aggregator SumAggregator = (blocks, primitiveSupplier) => blocks.Select(primitiveSupplier).Aggregate((Primitive)null, (a, b) => a?.Plus(b) ?? b) ?? ResolvePrimitive(0);
```
MerlinofMines, line 23: Why was this change needed?

EasyCommands/Common/Properties.cs
```diff
@@ -45,19 +45,19 @@ public class PropertySupplier {
...
-            PropertySupplier ResolvePropertyType(BlockHandler blockHandler, Return? defaultType = null) {
+            PropertySupplier ResolvePropertyType(IBlockHandler blockHandler, Return? defaultType = null) {
                 if (propertyType != null) return this;
-                if (direction.HasValue) return blockHandler.GetDefaultProperty(direction.Value);
-                if (propertyValue != null) return blockHandler.GetDefaultProperty(propertyValue.GetValue().returnType);
-                if (defaultType.HasValue) return blockHandler.GetDefaultProperty(defaultType.Value);
+                if (direction != null) return blockHandler.GetDefaultProperty(direction.Value);
+                var returnType = propertyValue?.GetValue().returnType ?? defaultType;
+                if (returnType != null) return blockHandler.GetDefaultProperty(returnType.Value);
                 return blockHandler.GetDefaultProperty(blockHandler.GetDefaultDirection());
```
MerlinofMines, line 61: nit:  

```
return returnType != null
? blockHandler.GetDefaultProperty(returnType.Value);
: blockHandler.GetDefaultProperty(blockHandler.GetDefaultDirection());
```

EasyCommands/Common/Utilities.cs
```diff
@@ -30,10 +30,16 @@ partial class Program {
         public static List<T> NewList<T>(params T[] elements) => new List<T>(elements);
         public static Dictionary<T, U> NewDictionary<T, U>(params KeyValuePair<T, U>[] elements) => elements.ToDictionary(e => e.Key, e => e.Value);
         public static KeyValuePair<T, U> KeyValuePair<T, U>(T key, U value) => new KeyValuePair<T, U>(key, value);
-        public static Variable EmptyList() => GetStaticVariable(NewKeyedList());
+
+        //Utilities for constructing enumerables
+        public static IEnumerable<int> Range(int start, int end) => Enumerable.Range(start, end);
+        public static IEnumerable<T> Empty<T>() => Enumerable.Empty<T>();
+        public static IEnumerable<T> Once<T>(T element) => Enumerable.Repeat(element, 1);
```
MerlinofMines, line 37: not sure I understand the value add from these ones.  How much better is the performance?  These two cost around 100 chars.

nit: can we remove ```Enumerable.```, as we did elsewhere?

EasyCommands/BlockHandlers/TextSurfaceHandlers.cs
jgersti: no allocation that gets thrown away immediatly since that expression is expanded in a `SelectMany` anyway.

EasyCommands/BlockHandlers/TextSurfaceHandlers.cs
jgersti: see above

EasyCommands/CommandParsers/ParameterParsers.cs
jgersti: see above

EasyCommands/Common/Utilities.cs
jgersti: if i havent miscounted it is the same right now.

EasyCommands/Commands/Commands.cs
jgersti: :)

EasyCommands/Commands/Commands.cs
jgersti: `null > 0` is `false` so giving a sensible default is not necessary here.

EasyCommands/Common/Aggregators.cs
jgersti: it saves chars. it may also enable us to deliver a appropriate default value vor non numeric inputs in the future.

EasyCommands/Common/Utilities.cs
jgersti: The main purpose of these functions is to better signal intent and avoidance of allocations. Some but not all usage can be replaced with `NewList`. All in all adding these functions costs only around 20-30 chars but increases readability and removes allocations.

Instead of `Enumerable.` these functions are called, so no

EasyCommands/BlockHandlers/TextSurfaceHandlers.cs
jgersti: also the branches inside `?:` must have the same type

EasyCommands/CommandParsers/ParameterParsers.cs
```diff
@@ -402,31 +402,27 @@ partial class Program {
...
-                commandParameters.Add(new AmbiguousStringCommandParameter(token.original, false, subtokenParams.ToArray()));
-            } else if (propertyWords.ContainsKey(t)) {
-                commandParameters.AddList(propertyWords[t]);
-            } else { //If no property matches, must be a string
+            else if (token.isString)
+                commandParameters.Add(new AmbiguousStringCommandParameter(token.original, false, ParseCommandParameters(Tokenize(token.token)).ToArray()));
+            else if (propertyWords.ContainsKey(token.token))
+                commandParameters.AddList(propertyWords[token.token]);
+            else //If no property matches, must be a string
                 commandParameters.Add(new AmbiguousStringCommandParameter(token.original, true));
```
MerlinofMines, line 425: nit: I think we can do the following:

```
            if (token.isExplicitString)
                commandParameters.Add(new VariableCommandParameter(GetStaticVariable(token.original)));
            else if (!token.isString && propertyWords.ContainsKey(token.token))
                commandParameters.AddList(propertyWords[token.token]);
            else 
                commandParameters.Add(new AmbiguousStringCommandParameter(token.original, token.isString, ParseCommandParameters(Tokenize(token.token)).ToArray()));
```

MerlinofMines: not sure if the extra tokenization matters.  *shrug*

EasyCommands/CommandParsers/ParsingEngine.cs
```diff
@@ -40,30 +40,16 @@ partial class Program {
...
-                }
+                    var nameAndParams = Tokenize(commandStrings[i].Remove(0, 1).Trim());
+                    var functionName = nameAndParams[0].original;
 
-                foreach (int i in functionIndices) {
-                    int startingLineNumber = i + 1 + implicitMainOffset;
-                    String functionString = commandStrings[i].Remove(0, 1).Trim();
-                    List<Token> nameAndParams = Tokenize(functionString);
-                    String functionName = nameAndParams[0].original;
+                    functions[functionName] = new FunctionDefinition(functionName, nameAndParams.Skip(1).Select(t => t.original).ToList());
```
MerlinofMines, line 49: lol glad to have removed this.  i think it was a relic from before dynamic function parsing, where i needed to know the function names up front and so had to do 2 passes.  Glad to simplify this.

MerlinofMines (approved):

PR #225, opened 2022-02-24
Small Changes

jgersti:
Some more char savings.

Keep the PR open to collect some more _small_ changes


EasyCommands/CommandParsers/ProcessingEngine.cs
```diff
@@ -432,15 +426,12 @@ partial class Program {
...
-            foreach (IParameterProcessor processor in processors) {
-                sortedParameterProcessors[processor] = processor.Rank;
-                processorRanks.Add(processor.Rank);
-            }
+        void AddProcessors(List<IParameterProcessor> processors, List<ICommandParameter> types) {
+            processors.AddRange(types
+                .Select(t => t.GetType())
+                .SelectMany(t => parameterProcessorsByParameterType[t])
+                .Except(processors));
+            processors.Sort();
```
MerlinofMines, line 434: I worry that our processing efficiency will suffer if we re-sort the processors after every processor insertion.   Unless my logic is faulty, the current implementation is log n for adding the ranked parameter processors into the sorted list.

This method will likely be n log n.

jgersti: ### `HashSet` vs `Except`
The iterator for `Except` internally holds a `Set`, so both imlementations have a complexity of `O(n+m)` for construction + filtering. Difference is the set internal to `Except` is rebuild every time and uses the default (object) comparison instead of the default (int) hash function. 

### `List`+`Sort` vs `SortedList`
`List` insertion is `O(1)` per element for a total of `O(k)`. The sorting algorithm has an average complexity `O(n log n)`. Element access is `O(1)` and removal `O(n)`. 

Direct `SortedList`  access is `O(log n)` for elements already contained and `O(n)` for new elements. Since we are prefiltering the elements to add total complexity for insertion is `O(k n)`.  Element access into the internal key array is again `O(1)` and element removal is `O(log n) + O(n)` (the first term can be removed by using the index that is known).
All modification operations on `SortedList` are twice as expensive since it keeps two internal arrays (`keys` and `values`).

If the default comparer is used both `List` and `SortedList` use a C++ implementation for the sort algorithm resp. the binary search.

### TL;DR
`SortedList` is worse than `List` in our case since we batch insert the elements and are limited to only a small number of elements (there are only 76 rules right now).

MerlinofMines: Not sure if it will matter to our performance, but I do believe the "Except" is adding O(n) complexity, as the HashSet was previously only constructed once, and then modified using O(1) check, removal and additions.  

I agree that List is going to be as, or more, efficient than SortedList.  I was intending behavior closer to SortedSet, which has it's own set of problems :).  Seems like List should be fine.  

It'd be good to do a quick performance sanity test on this though.   If this change were to cause a significant performance impact it would potentially affect dozens of folk's scripts, which would be very bad indeed.  Can you sanity test this before we merge, pretty please?

Thanks!

jgersti:
reordered commits for testing impact

EasyCommands/CommandParsers/ProcessingEngine.cs
```diff
@@ -432,15 +426,12 @@ partial class Program {
...
-
-            foreach (IParameterProcessor processor in processors) {
-                sortedParameterProcessors[processor] = processor.Rank;
-                processorRanks.Add(processor.Rank);
-            }
+        void AddProcessors(List<IParameterProcessor> processors, List<ICommandParameter> types) {
+            processors.AddRange(types
+                .Select(t => t.GetType())
+                .SelectMany(t => parameterProcessorsByParameterType[t])
+                .Except(processors));
```
MerlinofMines, line 433: nit: Future char saving:

```
.Select(t => parameterProcessorsByParameterType[t.getType()]
```

MerlinofMines (approved):

PR #230, opened 2022-03-17
Do not leak `PROGRAM`

jgersti:
only set `PROGRAM` during the constructor and `Main()` and make sure it is unset in a `finally` block. 
_This is necessary since static variables inside the script class are leaked. In our case this means `PROGRAM` kept the whole script and potentially cached blocks from being garbage collected._

The tests not using the `ScriptTest` mechanism are adapted to set `PROGRAM` appropriately.
also includes changed exceptions and some additional char savings.

### DIFF
--- a/EasyCommands/Program.cs
+++ b/EasyCommands/Program.cs
@@ -151,7 +151,7 @@ namespace IngameScript {
                 Info(stateTuple.Key);
                 Runtime.UpdateFrequency = stateTuple.Value ? updateFrequency : UpdateFrequency.None;
             } catch (Exception e) {
-                Print($"{(e is ParserException ? "Parser Exception" : e is RuntimeException ? "Runtime Exception" : "System Exception")} Occured:");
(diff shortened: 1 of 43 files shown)


EasyCommands/Program.cs
```diff
@@ -149,10 +150,13 @@ public partial class Program : MyGridProgram {
                 var stateTuple = programStateMap[state];
                 Info(stateTuple.Key);
                 Runtime.UpdateFrequency = stateTuple.Value ? updateFrequency : UpdateFrequency.None;
-            } catch(Exception e) {
-                Print("Exception Occurred:");
+            } catch (Exception e) {
+                Print($"{(e is ParserException ? "Parser Exception" : e is RuntimeException ? "Runtime Exception" : "System Exception")} Occurred:");
                 Print(e.Message);
```
MerlinofMines, line 155: nit: I might suggest "Parsing Exception" vs "Parser", and "Unknown" vs "System".  Not sure users will know the difference between Runtime & System.

PR #231, opened 2022-03-20
[Bugfix] Function Request For Updated Request

MerlinofMines:
Fixed issue where requesting a specific function to be executed after re-parsing did not work correctly.  Previously the main function would be executed after parsing completed, regardless of whether the user initiated a re-parsing of the script by changing the script on a non-running program, and then requested a specific function to run (either by a button click or manually requesting from the terminal menu).

### DIFF
--- a/EasyCommands/Program.cs
+++ b/EasyCommands/Program.cs
@@ -132,6 +132,8 @@ namespace IngameScript {
 
                 if (!ParseCommands()) {
                     Runtime.UpdateFrequency = updateFrequency;
+                    state = ProgramState.RUNNING;
+                    QueueRequest(argument);
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


EasyCommands/Program.cs
```diff
@@ -132,6 +132,8 @@ public partial class Program : MyGridProgram {
 
                 if (!ParseCommands()) {
                     Runtime.UpdateFrequency = updateFrequency;
+                    state = ProgramState.RUNNING;
+                    QueueRequest(argument);
```
jgersti, line 136: What happens if `ParseCommand` has not finished parsing yet?

MerlinofMines: If so, then the passed in request gets queued for execution and will be executed first (or, in queued order) once parsing is complete.

PR #236, opened 2022-03-31
Added Grid Block Handler

### LINKED ISSUE #222: [Feature Request] Grid BlockHandler
**Describe the solution you'd like**
We should add support for getting/setting values from the grid, such as the grid name, grid size, whether or not it is static, etc.

I suggest a "GridHandler" that returns ProgrammableBlock.Me.CubeGrid.  Can have properties for the size, name, and maybe static?

MerlinofMines:
This commit adds basic support for Grids, providing access to get/set the grid name, and read-only access to whether the grid is static and the grid size
(large/small).

In addition, this commit removes the default String "name" property, which wasn't providing much benefit and usually caused confusion.  The new expectation is
that if you want to set a block's name, you must specify the "name" property.

This PR fixes #222


jgersti:
Having a plural from and mentioning multiple grids in the documentation is misleading, we can only affect the whole grid the PB is on. 
Even if we could discriminate between the subgrids we have no properties right now that would differ between the subgrids.

MerlinofMines:
> Having a plural from and mentioning multiple grids in the documentation is misleading, we can only affect the whole grid the PB is on. Even if we could discriminate between the subgrids we have no properties right now that would differ between the subgrids.

There are multiple grids present on which the PB relies.  All sub grids are considered separate grids, so you can actually do things like ```set gridCount to the count of grids```  As such I think the plural form is reasonable, though probably won't be used in most cases.

PR #237, opened 2022-03-31
Remove `IgnoreCommandParameter`

jgersti:
This PR introduces the following changes:
 *  Introduce new command parameters for `to` (`AbsoluteCommandParameter`) and `in` (`InCommandParameter`).
    All other tokens that are currently ignored can then be discarded before even invoking the actual parser.
    This allows removal of the rule that discards `IgnoreCommandParamter` and `IgnoreCommandParameter` itself.
 *  Modify the rule for `VariableIncrementParameter` to include a optional `RelativeCommmandParameter` (`by`).
    This allows removal of the rule that discards `RelativeCommandParameter`.
 *  Modify the rule for `IteratorComandParameter` to include a optional `InCommandParameter`.
     Modify the rules that produce `PropertySupplierCommandParameter`s to include a optional `InCommnadParameter` (`in`) to the left in case of the rule for `PropertyCommandParameter` and to the right in case of the rules for `ValuePropertyCommandParameter`.
    This handles all cases where `in` is used inside the test.
 *  ~~Add a rule to discard `AbsoluteCommandParameter` before the `BlockCommandParameter` rule.  This is a temporary measure to keep everything running for now . To resolve this the `BlockCommandParameter` rule most likley needs to be split into seperate cases for assignment and increment/increase.
    All the assignment rules either before or after the `BlockCommandParameter` need to be changed to consume a `AbsoluteCommandParameter`~~
    Add `AbsoluteCommnadParameter`s to various rules and do not discard it as a standalone CP.

Changes not yet included are:
* ~~Introduce a `IdentifierCommandParameter` to simplify/remove `CanConvert` with checks like `name.GetValue().value is AmbiguousStringVariable`.~~ _Added. see below_
* Replacement of `ParenethesisProcessor`, `ListProcessor` and `MultiListProcessor` with conventional rules. This might need some changes to the parser itself like a non consuming lookahead processor.
* Match processors for a rules in order of the defintion instead of unorderd. Currently any rule that matches more than one processor to the left or to the right can match these in any order and the preferred order of the left matches is also reversed. Atleast `BlockCommandProcessor` and the rules for item transfer (?) depend on this behaviour.
   As a first step add a paramter to `RuleProcessor` that allows to select the desired behaviour and determine how many rules depend on the old behaviour.

_As a side note :
I would propose to rename `*CommandParameter => *Token`, `*DataProcessor => *Match` and the current `Token => Lexeme`.
And maybe also renaming a few files to reflect what they contain `ParameterParser.cs => Lexer.cs`, `ProcessingEngine.cs => Parser.cs`_
     
_edit: 
07. April: updated some points above_

### DIFF
--- a/EasyCommands/CommandParsers/ParameterParsers.cs
+++ b/EasyCommands/CommandParsers/ParameterParsers.cs
@@ -67,7 +67,7 @@ namespace IngameScript {
 
         public void InitializeParsers() {
             //Ignored words that have no command parameters
-            AddWords(Words("the", "than", "turned", "block", "panel", "chamber", "drive", "to", "from", "then", "of", "either", "for", "in", "do", "does", "second", "seconds", "be", "being", "digits", "digit"), new IgnoreCommandParameter());
+            AddWords(Words("the", "than", "turned", "block", "panel", "chamber", "drive", "from", "then", "of", "either", "for", "do", "does", "second", "seconds", "be", "being", "digits", "digit"));
 
             //Selector Related Words
             AddWords(Words("blocks", "group", "panels", "chambers", "drives"), new GroupCommandParameter());
@@ -96,6 +96,7 @@ namespace IngameScript {
             AddWords(Words("--", "-="), new IncrementCommandParameter(false));
             AddWords(Words("global"), new GlobalCommandParameter());
             AddWords(Words("by"), new RelativeCommandParameter());
+            AddWords(Words("to"), new AbsoluteCommandParameter());
 
             //Value Words
             AddWords(Words("on", "begin", "true", "start", "started", "resume", "resumed"), new BooleanCommandParameter(true));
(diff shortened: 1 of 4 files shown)


jgersti:
___Move the detection of `Primitive`s into `ParseCommandParameter`___
This commit disallows the usage of tokenized primitves as selectors and function names.
The changed behaviour should only be noticeable if the predefined color names where used since hex colors, vectors and numbers should not be used as identifiers anyway.

jgersti:
___Add `AbsoluteCommandParameter`s to various rules and ...___
This commit removes the rule that discards `AbsoluteCommandParameter`. Any rule that needs to be aware of this has been changed to include an data processor that reads `AbsoluteCommandParameter`.

This commit is more a quick fix and not proper solution since `BlockCommandProcessor` now matches even more patterns (50+ million assuming only up to 2 assignment matches).

MerlinofMines:
Multi-Word property support has some conflicts with this one, going in a different direction in order to support "To" separation.

closing this draft for now; we can revisit later.

